PR #68, opened 2016-07-27
Update mil-fujifilm.xml

karlstevens:
Added entry for XF16-55mmF2.8 R LM WR from lens data b73a4c (issue #56)


bronger:
Please thin out the focal lengths.  5 are enough I think, closer at the wide-angle end (see [tutorial](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/calibration.rst#distortion)).  Note that double focal lengths are not allowed.

Issue #77, opened 2016-07-30
Calibration upload e6a211


bronger:
Calibration images were uploaded to the directory that starts with “`e6a211_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

geneing:
Please note that I uploaded images taken with both Olympus 40-150 Pro lens and with the same lens+MC-14 teleconverter. 
EXIF lens name information is different for both. They probably should be treated as 2 different lenses.

bronger:
Indeed, they must be two different lenses for lensfun.

bronger:
> I also did the corrections for the 1.4x teleconvertor, but for some reason darktable is refusing to apply those corrections, even when selected manually (and shows the focal range as 40-210mm, weird). I also wasn't sure how it should be named, exactly.

I think naming is important here.  See https://github.com/lensfun/lensfun/blob/master/libs/lensfun/lens.cpp#L246.  One of the words "adapter", "reducer", etc. must occur in the lens model name.  Then, Lensfun doesn't try to guess the focal length range anymore, and you can use 56 to 210mm without problems.

junkyardsparkle:
@bronger: Ah, those are the magic words I needed to know, thanks! Maybe a small section on naming in the HOWTO could include this and other non-obvious useful info?

junkyardsparkle:
@geneing: Although there is additional EXIF data for the teleconverter, I don't know if it's possible for lensfun in its current form to make use of it for auto-selection; as it is now, you'll need to manually select the lens for that case.

bronger:
Indeed, such adapters probably always need manual selection.  Even if the data is in the RAW, I don't think that exiv2 can make use of it.

bronger:
@junkyardsparkle @lensfun/calibrators I added a paragraph about what is allowed/necessary when changing lens model names to https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/calibration.rst#create-lensfun-data.  (Beginning with “Often, you have to fine-tune the lens model name”.)

sarunasb:
Thank you, that is very useful to know.

geneing:
Thank you very much for creating calibration files.

PR #86, opened 2016-08-06
Calibration data for Canon 100L Macro and 300 f/4

gnl21:
Add lenses from c70751 (issue #45)

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -3267,6 +3267,18 @@
 
     <lens>
         <maker>Canon</maker>
+        <model>Canon EF 100mm f/2.8L Macro IS USM</model>
+        <mount>Canon EF</mount>
+        <cropfactor>1.605</cropfactor>
+        <calibration>
+            <!-- Taken with Canon 7D Mark II -->
+            <distortion model="ptlens" focal="100" a="0.00804" b="-0.0341" c="0.04884" />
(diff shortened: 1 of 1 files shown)


data/db/slr-canon.xml
```diff
@@ -4665,6 +4676,28 @@
...
+        <cropfactor>1.605</cropfactor>
+        <calibration>
+            <distortion model="ptlens" focal="300" a="-0.00271" b="0.00945" c="-0.00645" />
+            <tca model="poly3" focal="300" br="0.0000114" vr="0.9999147" bb="-0.0000121" vb="0.9999900" />
+        </calibration>
+    </lens>
+
+    <lens>
+        <maker>Canon</maker>
+        <model>EF300mm f/4L IS USM +1.4x</model>
```
bronger: Please append the word "converter".  See https://github.com/lensfun/lensfun/issues/77#issuecomment-236888725.

gnl21: Thanks, that's what I was hoping for feedback on. I thought I'd read through the calibration document, but clearly not well enough, sorry. I'll add the camera comments as well.

A Lot of the similar calibrations have translations for the converters as well, is it useful to copy that here?

gnl21:
I've updated the original commit to fix the problems identified. The second commit changes the lens names to be more consistent. I'm having some problems testing the teleconverter data in Darktable, which doesn't seem to be doing any correction at all.

Issue #83, opened 2016-08-02
Calibration upload 5d9390


bronger:
Calibration images were uploaded to the directory that starts with “`5d9390_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Granjow:
(Uploader is me.)

gnl21:
@Granjow @bronger I don't think these images are suitable for calibration for the 85mm lens. The TCA pictures are great but I think that the lens exhibits longitudinal CA, which I don't think can be corrected by lensfun.

In the distortion pictures I'm not sure that the diagonal lines will work (the ones that go near the corners are very short) and the drain-pipe may not actually be straight.

I can make a TCA calibration for the 70mm macro.

Granjow:
Thanks for the hint. Sorry for the bad quality. I now have found a better subject for calibration pictures and will upload a sample so you can check if it would work for the 85 mm lens.

gnl21:
No problem. I've uploaded TCA data for the 70mm macro. If you upload the new shots then I'll have a look at making a calibration for the 85 as well.

Granjow:
Does this one work better?
http://granjow.net/uploads/temp/DSC_3740.NEF

gnl21:
Looks like I pushed the TCA data to the wrong place, but it should be uploaded now. The new image looks great, so I'll do a calibration for it.

Granjow:
Thanks!

bronger:
Are there two different 85mm in the set?  My Exiftool says “Sigma 85mm F1.4 EX DG HSM” for some, and “85mm f/1.4” for the others.

bronger:
While the idea with the balls is great, the images are overexposed and not always sharp, which is difficult for `tca_correct`.  I think one should use `tca_correct` with the `--simple-tca` option because TCA is weak here, and the images not optimal.  That said, the current data definitely makes an improvement, and it may be as good as it gets.

As far as distortion is concerned: If really no distortion can be measured (happened also to me once) please use the poly3 model as its maths are simpler and don't need as much CPU.

Granjow:
It should all be the same lense because I just have this one. But something just comes into my mind, I did not take the most recent picture at f/1.4, but at f/2.2. I assume you need a f/1.4 one?

gnl21:
[...]

PR #106, opened 2016-09-19
Fix simple-tca handling in calibrate.py

gnl21:
Only switch on simple-tca if it is requested. Passing no arguments will
now select the full model.

### DIFF
--- a/tools/calibrate/calibrate.py
+++ b/tools/calibrate/calibrate.py
@@ -265,7 +265,7 @@ except IOError:
 # TCA correction
 #
 
-only_linear_tca = len(sys.argv) <= 1 or sys.argv[1] == "--simple-tca"
+only_linear_tca = len(sys.argv) > 1 and sys.argv[1] == "--simple-tca"
 
 def generate_tca_tiffs(filename):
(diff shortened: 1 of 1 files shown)


bronger:
Why do you think the complex model is the better default?

bronger:
Is related to https://github.com/lensfun/lensfun/issues/83#issuecomment-246188207.

gnl21:
Only because the option is called 'simple -tca', so not passing the option
should give the opposite of that. Which should be the default was what I
was asking about. Currently no option gives you simple but passing
'simple-tca' changes to the full model.

gnl21:
Actually, I think I'm wrong in the last comment. Passing 'simple-tca' does nothing, passing anything else switches to the full model.

bronger:
How embarassing … it is now fixed in master.  I dropped `--simply-tca` and introduced `--complex-tca`.  Simple TCA remains the default.  And, there is no fix for the problem that if you change the TCA model, you must delete the temporary files in `tca/` first. (I understood this problem correctly, right?)

Thank you for the report!

@lensfun/calibrators

PR #123, opened 2016-10-23
compact-canon.xml: Add full A610 profile for CHDK DNGs

hknaack:
Full profiling (distortion, tca, vignetting) done on a PowerShot A620 (same lens as A610) using CHDK DNG files. This covers all 9 zoom steps, which can be selected in CHDKs uBasic.


bronger:
Thank you very much for the data!

PR #124, opened 2016-10-29
Fix BUILD_TESTS: Pull isnan() into std:: namespace, #include <cmath>, not <math.h>.

mandree:


bronger:
We include math.h in many places.  Should they all be changed?

mandree:
This is the only one that cause compiler trouble on my system, and only in the patched file, so "no need to touch others". To reproduce, enable BUILD_TESTS with ccmake and compile.

bronger:
I merged this, although I cannot see all implications, especially for Windows.  Please report bugs at lensfun.sourceforge.net.  The GitHub repository is only for coordinating lens calibations.  Thank you!

PR #135, opened 2016-12-09
Update mil-panasonic.xml

sarunasb:
Add Panasonic GX80/85 <camera> elements (DMC-GX80 and DMC-GX85).


bronger:
Unfortunately, I looked at this PR too late.  It is already implemented.

PR #173, opened 2017-02-21
Add Nikon D5 and Nikon D500 SLRs

bstegmaier:
D5 sensor size: 35.9mm x 23.9mm
D500 sensor size: 23.5mm x 15.7mm


bronger:
Thank you very much!

PR #180, opened 2017-03-26
Add full data for Sony FE 85mm F1.8 (SEL-85F18).

mandree:


mandree:
I just figured that the camera bakes in Vignetting compensation into the RAW, and these data were obtained with this feature turned on. I'll need to create alternate data sets for a few lenses...

PR #201, opened 2017-05-30
Correct some Pentax crop factors

gnl21:
The crop factors for the Pentax K-01 and KP appear to have been slightly wrong. As reported by Julien Moreau, and confirmed from specs on the internet.


bronger:
Have lenses been calibrated using the affected sensors?

gnl21:
The K-01 was only used to calibrate the lens that is also updated in the commit (the crop factor only appears in those two entries in the DB). I don't think that the KP was used in any calibrations -- no lenses say they were and I don't see an entry with that crop factor that doesn't say what it was calibrated with.

Issue #263, opened 2017-11-04
Calibration upload 08f37b


bronger:
Calibration images were uploaded to the directory that starts with “`08f37b_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Vignetting data of:
> AF Nikkor 180mm f/2.8D IF-ED
> 
> taken with a Nikon D3

jakobdettner:
I am the uploader.

PR #274, opened 2017-11-05
Vignetting nikon 180mm 2.8 d

jakobdettner:
resolved issue #263


jakobdettner:
oups, added also another lens (60mm/2.8) which I already created a pull request for. Is this a problem?

matt-maguire:
It looks like these correcttions were already added to the database by some other means. Thanks for the contribution!

jakobdettner:
It would be interesting if the two lens corrections match each other.

Issue #271, opened 2017-11-04
Calibration upload 982ef2


bronger:
Calibration images were uploaded to the directory that starts with “`982ef2_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Vignetting data of:
> 
> Nikkor AI-S 135mm f/2.8
> 
> taken with a Nikon D3

jakobdettner:
I am the uploader.

Issue #267, opened 2017-11-04
Calibration upload d4ff18


bronger:
Calibration images were uploaded to the directory that starts with “`d4ff18_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Vignetting data of:
> Nikkor AI-S 50mm f/1.8
> 
> taken with a Nikon D3

jakobdettner:
I am the uploader.

Issue #266, opened 2017-11-04
Calibration upload e968d0


bronger:
Calibration images were uploaded to the directory that starts with “`e968d0_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Vignetting data of:
> Nikkor AI-S 50mm f/1.4
> 
> taken with a Nikon D3

jakobdettner:
I am the uploader.

Issue #265, opened 2017-11-04
Calibration upload cc506a


bronger:
Calibration images were uploaded to the directory that starts with “`cc506a_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Vignetting data of:
> Nikkor AI 20mm f/3.5
> 
> taken with a Nikon D3

jakobdettner:
I am the uploader.

PR #275, opened 2017-11-06
Add nikon slr vignetting data

jakobdettner:
resolves: #271 , #267 , #266 and #265 

Also added vignetting data of AF Nikkor 35mm f/2D and AF Zoom-Nikkor 35-70mm f/2.8

For the zoom lens I used 7 different focal lengths. Is that ok or should it be reduced to 5?


jakobdettner:
What are the further steps to get this into master? 
Is there something I should do or what would be the process?

Issue #191, opened 2017-04-23
Calibration upload 16eff9


bronger:
Calibration images were uploaded to the directory that starts with “`16eff9_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Some of these lenses you already have, but with a K1 body they now should have a crop factor of 1.0
> 
> Thanks!

bronger:
@gnl21 still working on this?

gnl21:
Sorry, I'd forgotten about this. I did the calibration and it looked good, but when I tested it it didn't work at all. I'll sort it out in the next day or two.

gnl21:
Actually, it seems pretty straightforward now. The camera (Pentax K-1) isn't in the database, which prevented me from selecting the lens data that I'd added.

gnl21:
So I've sorted out 2 of the 3 lenses in this issue. The Pentax K-1 is in the git copy of the database but isn't included in the database on the production machine that I use for testing. Running lensfun-update-data says that there are no updates for that machine, is that correct? Either way, adding the camera locally the calibrations look good.

The third lens is a crop-camera lens but the images are from a full-frame camera. Should things like that be included in the database? The TCA in the images also appears to be somewhat asymmetric (the calibration works well on the right side of the frame but are less good on the left). If you think it's worth continuing with that third lens then I'll finish it off.

gnl21:
There's the two lenses that are done. I'll finish this off if you let me know what to do about the adapted lens @bronger.

karlstevens:
Is it an adapted lens, or just one with a reduced image circle?

gnl21:
I meant adapted in the sense of 'used on a mount with which it's not nominally compatible', I don't know whether this combination would require a physical adapter. It's a Pentax APS-C lens on a full frame Pentax camera.

karlstevens:
Pentax K-mount lenses are nominally compatible with all Pentax bodies.  Lenses designed for APS-C bodies may experience vignetting (ranging from a stop or two, down to full occlusion in the corners) when used on full-frame bodies, and the K-1 allows the user to select the APS-C area or the full sensor area when taking photos.   (The reasoning behind this is that the image area will always extend beyond the APS-C sensor area.)

Since we have many lenses that do not cover the full sensor (the Leica Q being the most recent that comes to mind), I think it should be included.

[...]

PR #281, opened 2017-11-09
Add calibration data for Miranda 28/2.8

gnl21:
Issue #191, calibration 16eff9

### DIFF
--- a/data/db/misc.xml
+++ b/data/db/misc.xml
@@ -483,4 +483,21 @@
         </calibration>
     </lens>
 
+    <lens>
+        <maker>Miranda</maker>
+        <model>Miranda 28mm f/2.8 MC</model>
+        <mount>Contax/Yashica</mount>
+        <mount>Canon FD</mount>
+        <mount>M42</mount>
+        <mount>Nikon F</mount>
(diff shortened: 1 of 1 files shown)


data/db/misc.xml
```diff
@@ -483,4 +483,21 @@
         </calibration>
     </lens>
 
+    <lens>
+        <maker>Miranda</maker>
```
gnl21: I wasn't totally sure where to add this. The internet suggests that this might be the same as the Cosina 28mm f/2.8, but I'm not sure.

data/db/misc.xml
```diff
@@ -483,4 +483,21 @@
...
+        <model>Miranda 28mm f/2.8 MC</model>
+        <mount>Contax/Yashica</mount>
```
gnl21: This is the list of mounts that came up when Googling this lens, but I don't know how to verify it. Is there a canonical (or at least reliable) source?

data/db/misc.xml
bronger: Let's take what it says on the lens housing.  Apparently Miranda, accorgings to the uploader …

PR #293, opened 2017-11-27
Add calibration for Canon EF-S 55-250mm f/4-5.6 IS II at crop 1.613

HeRoSofty:
Taken with Canon EOS 77D and processed with calibration.py

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -2288,6 +2288,88 @@
 
     <lens>
         <maker>Canon</maker>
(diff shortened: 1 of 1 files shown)


matt-maguire:
We already have a correction profile in the DB for this lens, reusing the profile from the Mark I model (optics are supposed to be identical in the Mark II, just some external casing improvements).

The profile in this PR contains corrections for an extra focal length, so it is perhaps worth comparing the existing profile with this one proposed in this PR to determine which profile is best to proceed with.

tuxfanx:
4 years old  and no complaints about the existing profile --> closed

PR #435, opened 2018-06-27
Added Vignette correction for Nikon DX lenses

hrlmartins:
Added Vignette correction to the following nikon DX lenses:
Nikon AF-S DX Nikkor 18-140mm f/3.5-5.6G ED VR
Nikon AF-P DX Nikkor 70-300mm f/4.5-6.3G ED VR

Tested with the images used for calibration and on field footage and compared to previously edited ones in lightroom.

The cropfactor of my camera is slightly different for the one that already exists for the 18-140mm lens (1.534 instead of 1.523). But to make it work locally on the local database (~/local/share/lensfun/ path) on darktable, I had to put it under the 1.523 crop one. Not sure if it is intended just FYI :)


FreddieWitherden:
I think that the corrections for Nikon AF-P DX Nikkor 70-300mm f/4.5-6.3G ED VR can also be applied to the non-VR model since the optics are very similar.

PR #472, opened 2018-09-02
Added calibration parameters for Meike 25mm f/1.8

manu-mannattil:
Meike 25mm f/1.8 is a lens designed for APS-C mirrorless cameras.  This
commit adds distortion, TCA, and vignetting correction for the lens.

### DIFF
--- a/data/db/misc.xml
+++ b/data/db/misc.xml
@@ -438,6 +438,38 @@
 
     <lens>
         <maker>Meike</maker>
+        <model>Meike 25mm f/1.8</model>
+        <mount>Canon EF-M</mount>
+        <mount>Fujifilm X</mount>
+        <mount>Micro 4/3 System</mount>
+        <mount>Nikon CX</mount>
+        <mount>Sony E</mount>
+        <cropfactor>1.534</cropfactor>
+        <calibration>
+            <!-- Taken with Sony ILCE-6000 -->
+            <distortion model="ptlens" focal="25" a="0.023" b="-0.067" c="0.05"/>
+            <tca model="poly3" focal="25" vr="1.0006443" vb="0.9995646"/>
+            <vignetting model="pa" focal="25" aperture="1.8" distance="10" k1="-0.6979" k2="-0.5158" k3="0.4266"/>
+            <vignetting model="pa" focal="25" aperture="1.8" distance="1000" k1="-0.6979" k2="-0.5158" k3="0.4266"/>
+            <vignetting model="pa" focal="25" aperture="2" distance="10" k1="-0.6704" k2="-0.4621" k3="0.3552"/>
+            <vignetting model="pa" focal="25" aperture="2" distance="1000" k1="-0.6704" k2="-0.4621" k3="0.3552"/>
+            <vignetting model="pa" focal="25" aperture="2.8" distance="10" k1="-0.4438" k2="-0.1696" k3="-0.0930"/>
+            <vignetting model="pa" focal="25" aperture="2.8" distance="1000" k1="-0.4438" k2="-0.1696" k3="-0.0930"/>
+            <vignetting model="pa" focal="25" aperture="4" distance="10" k1="-0.6661" k2="0.7509" k3="-0.7460"/>
+            <vignetting model="pa" focal="25" aperture="4" distance="1000" k1="-0.6661" k2="0.7509" k3="-0.7460"/>
+            <vignetting model="pa" focal="25" aperture="5.6" distance="10" k1="-0.6639" k2="0.6166" k3="-0.5545"/>
+            <vignetting model="pa" focal="25" aperture="5.6" distance="1000" k1="-0.6639" k2="0.6166" k3="-0.5545"/>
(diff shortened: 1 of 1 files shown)


jhol:
Heh, we added the same lens!: https://github.com/lensfun/lensfun/pull/958

I tested mine on M4/3

manu-mannattil:
Yeah, but the calibration parameters look different. Hopefully, that's due to the difference in crop factors?  This is the first time I'm calibrating a lens, so I might have made a mistake as well (it seems to work fine for me though).

jhol:
Yes... this is my first submission also. The crop factor is the only explanation I can think of for the differences.

It seems to me that your pull request should supersede mine given that you have a larger sensor and therefore you're using a wider area of the glass.

junkyardsparkle:
Yes, the different crop factors explain this. @jhol, can you try using this profile instead of yours, and confirm that the results are still good? If so, I'll commit this one, since it should cover both cases.

jhol:
I've been doing a comparison, and the results are not looking quite right.

Here is a thumbnail of my test image:

20191118-meike-25mm-f1 8-test-thumbnail

Here is the version corrected by my PR:

20191118-meike-25mm-f1 8-test-jhol-corrected-thumbnail

Here is the version corrected by this PR:

20191118-meike-25mm-f1 8-test-manu-mannattil-corrected-thumbnail

It looks weirdly over-exposed and more wonky than my version - though neither one is perfect.

Here are the unshrunk PNG files of the above test: https://drive.google.com/file/d/1NL3YxIBwXbjGv9QQ4C0jO4D5-aXm54If/view?usp=sharing

junkyardsparkle:
> It looks weirdly over-exposed and more wonky than my version - though neither one is perfect.

I'm going to guess that focus distance might be the reason; that target looks rather small. Distortion calibration for the database shouldn't be done at close distances, something like 10 meters is better as a general guideline.

manu-mannattil:
The calibration parameters from my PR seems to do more vignetting correction than @jhol's, which probably explains the overexposure (since an APS-C sensor is slightly larger than an M4/3 sensor, the same lens should introduce more vignetting).  In any case, for the test image, I got better results by manually changing the aperture to f/16.0. (Full-sized samples [here](https://drive.google.com/file/d/1x6ojPkMhCu5xoFEc4Zs4V8dD9-CvHJ0f/view?usp=sharing) and thumbnails below.)

I also wanted to compare the two calibrations using an image taken with a camera having an M4/3 sensor and a relatively far away object like @junkyardsparkle suggested.  I found [this image](https://www.flickr.com/photos/158567241@N07/30238960488/in/pool-3708210@N25/) taken using a Lumix DMC-GH4 + Meike 25mm f/1.8 from Meike's official Flickr account.  I'm guessing that this was taken at a low aperture given the overall exposure and the 1/200s shutter speed.

For this image, the differences between the two calibrations aren't too apparent.  However, I should note that my calibration parameters seem to do slightly more aggressive vignetting correction, which may result in an overexposed image if the aperture isn't set properly.  Vignetting correction using @jhol's parameters doesn't seem to be affected by aperture as much as mine.  @jhol's parameters seem to do more distortion correction though.  I think I should experiment with this lens a bit more before finalizing the parameters.

---

@jhol's test image using my parameters and setting the aperture to f/16.0:

xjhol-test-mm-f16 0

---

Flickr sample using @jhol's calibration parameters (manually setting the aperture to f/1.8):

xflickr-test-jhol

---

Flickr sample using my parameters (manually setting the aperture to f/1.8):

xflickr-test-mm-f1 8

---

Flickr sample using my parameters (manually setting the aperture to f/8.0):

xflickr-test-mm-f8 0

junkyardsparkle:
> The calibration parameters from my PR seem to do more vignetting correction than @jhol's,  which probably explains the overexposure (since an APS-C sensor is  slightly larger than an M4/3 sensor, the same lens should introduce more  vignetting). 

Lensfun should account for the different crop factors in its calculations (that's why profiles from larger sensors can also be used for smaller ones). The vignetting differences are probably focus distance related; all of the internally focusing m4/3 lenses I have vignette more strongly when focused at infinity than when focused close, which makes creating a good profile difficult. If you search the issues (open and closed) for vignetting + over-correction you'll find plenty of discussion about the problem.

LT;DR: My own approach for my personal profiles is to either do the profiling at many distances (much work, and only helps if the camera/lens combination give accurate distance in EXIF) or simply profile for the *least* case (usually closest focus, but maybe not for some lens designs, so should be checked). That way, the errors skew towards *under*-correction, which is much less objectionable than *over*-correction in general.

jhol:
Here's a picture of a brick wall taken at ~3m distance, at f1.8:

20191119-meike-25mm-f1 8-test-3m

junkyardsparkle:
@jhol Is that without corrections applied?

jhol:
Indeed... no corrections

junkyardsparkle:
Well, I applied this profile to that image, and igoring any vignetting issues as discussed above, it seems like a good fit to me, so unless anyone disagrees, I'll call it "tested" and add it to the database.

PR #495, opened 2019-01-10
Add Panasonic TZ202

bobobo1618:


bobobo1618:
Reviewing it with some out-of-sample images, it looks like this has issues.

junkyardsparkle:
Thanks for the data. There was a recently added profile for the ZS200, a different regional version of the same camera. Your data had had better coverage of the shorter focal lengths, so I updated the existing profile with that. FYI, fixed lens camera mount names should begin with a lowercase letter, as this is used for autoselection (by convention, 'panasonicTZxxx' format has been used on these cameras).

PR #496, opened 2019-02-21
[LENS] Add Sony SEL1635GM and SEL85F18GM

cryptomilk:
Hi,

this adds too new Sony lenses to lensfun.

You can find all the data here:

https://xor.cryptomilk.org/lensfun/SEL1635GM_calibration.tar.xz
https://xor.cryptomilk.org/lensfun/SEL85F18GM_calibration.tar.xz

Those files include the plots for TCA and vignetting. The package includes plots and preview files for vignetting corrections.

https://gitlab.com/cryptomilk/lens_calibrate

### DIFF
--- a/data/db/mil-sony.xml
+++ b/data/db/mil-sony.xml
@@ -1381,6 +1381,105 @@
 
     <lens>
         <maker>Sony</maker>
+        <model>FE 16-35mm f/2.8 GM</model>
+        <mount>Sony E</mount>
+        <cropfactor>1.0</cropfactor>
+        <calibration>
+           <distortion model="ptlens" focal="16" a="0.021" b="-0.052" c="-0.003" />
+           <distortion model="ptlens" focal="17" a="0.016" b="-0.033" c="-0.014" />
+           <distortion model="ptlens" focal="18" a="0.02" b="-0.05" c="0.009" />
+           <distortion model="ptlens" focal="19" a="0.023" b="-0.058" c="0.024" />
+           <distortion model="ptlens" focal="20" a="0.016" b="-0.036" c="0.004" />
+           <distortion model="ptlens" focal="21" a="0.027" b="-0.08" c="0.065" />
+           <distortion model="ptlens" focal="22" a="0.02" b="-0.057" c="0.046" />
+           <distortion model="ptlens" focal="23" a="0.018" b="-0.049" c="0.043" />
+           <distortion model="ptlens" focal="24" a="0.015" b="-0.036" c="0.027" />
+           <distortion model="ptlens" focal="25" a="0.019" b="-0.05" c="0.045" />
+           <distortion model="ptlens" focal="26" a="0.017" b="-0.0440" c="0.038" />
+           <distortion model="ptlens" focal="27" a="0.011" b="-0.027" c="0.029" />
+           <distortion model="ptlens" focal="28" a="0.01" b="-0.021" c="0.019" />
+           <distortion model="ptlens" focal="29" a="0.008" b="-0.012" c="0.008" />
+           <distortion model="ptlens" focal="30" a="0.013" b="-0.027" c="0.023" />
+           <distortion model="ptlens" focal="31" a="0.001" b="0.005" c="0.002" />
+           <distortion model="ptlens" focal="32" a="0.006" b="-0.013" c="0.026" />
+           <distortion model="ptlens" focal="33" a="0.003" b="0.0" c="0.01" />
+           <distortion model="ptlens" focal="34" a="-0.003" b="0.024" c="0.023" />
+           <distortion model="ptlens" focal="35" a="0.002" b="0.01" c="-0.009" />
+           <tca model="poly3" focal="16.0" vr="1.0003371" vb="1.0002001" />
+           <tca model="poly3" focal="17.0" vr="1.0002961" vb="1.0002093" />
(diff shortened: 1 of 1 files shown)


junkyardsparkle:
Thanks! Before adding this, could you confirm that the vignetting correction doesn't do any weird over-correcting on real-world (or non-diffuser flat-field) images? There's some discussion about that [here](https://sourceforge.net/p/lensfun/features/148/), for context. More likely to be an issue with the wider lens, probably.

cryptomilk:
I would argue it works just fine: https://pixelbook.org/#15392672842961

junkyardsparkle:
It very well might, but it doesn't hurt to check against some images specifically intended to show up any over-correction. Processed images aren't great for that, even if they're great pictures. :)

For example, [this one](https://pixelbook.org/#15392672842961/15429953210786) shows what *might* be over-correction, or it might just be the lighting or something else you did in processing. Anyway, to be clear, I'm *not* questioning your competence, it's just that the diffuser method of obtaining calibration images fails in some cases (I'm still trying to get a better grasp of which cases and why), so I'm just trying to encourage some extra checking for the profiles that go into the database. It's quite certain that some of the profiles already there suffer from this, but there's no way to test without the lens in hand. There's a recent case of somebody observing this problem [here](https://sourceforge.net/p/lensfun/bugs/113/).

cryptomilk:
Torsten wrote in his original tutorial, that you should direct a light to the ceiling and then take the pictures. The problem is that on the side where the light is located, the picture is brighter. You can then less correction on one side and possibly over-correction on the other side.

That's why I suggest in my tutorial [1] to shoot into the sky on a cloudy day. Then you get an evenly exposed shot. If the pictures are to bright you can user a paper to make it darker or in the other direction wait for a thinner cloud coverage.

If you have suggestions, add a comment to the article or modify it directly, the source code is on github.

[1] https://pixls.us/articles/create-lens-calibration-data-for-lensfun/

cryptomilk:
We could extend the calibration script to generate small thumbnails of the vignetting pictures and include them in the tarball generated at the end.

junkyardsparkle:
(Forgive me if this isn't clear, it's late here.) I'm still trying to find the best practical method for obtaining good vignetting images. In theory, a very flat, diffuse, featureless surface lit only by direct sunlight, at an angle enough to stand in front of it without casting a shadow, should be good for this, but there are probably things that can go wrong with that, like everything else.

Soon, when I have the time, I want to set up a large piece of white acrylic in a window of an otherwise unlit room, and take shots ranging from zero to ~1 meter distance, to find out what distance is actually needed to avoid the problems with the diffuser. After that I might have more useful suggestions.

For now the main thing is that whatever method is used, the results should be checked against real-world images (of sky or some plain surface) to avoid over-correction.  Thumbnails of the calibration images might be good for sanity-checking the images used... I've seen some dirty camera sensors that have a lot of "features" at f/22. ;)

cryptomilk:
But in direct sunlight you will have a shadows as you can't point the camera directly in the sun, on a cloudy day you can ...

Maybe we should move this discussion about how to take vignetting images to the pixls.us forums?

junkyardsparkle:
I mean using sunlight in a similar way to shooting a color calibration target, but the target is larger and devoid of features.

Forum discussions tend to have a poor signal/noise ratio in my experience, but start a discussion there if you think it will be productive. I need to do more empirical investigating before I do more discussing than I've already linked to.

cryptomilk:
Any news about your tests and findings?

cryptomilk:
Vignetting corrections in darktable and probably other tools don''t not work yet, as exiv2 doesn't provide a focal distance estimation for Sony yet. See https://github.com/Exiv2/exiv2/issues/582

I've already have a branch for darktable to fix it, but need support in exiv2 first. Probably later this year ...

junkyardsparkle:
> Vignetting corrections in darktable and probably other raw developers doesn't not work yet, as exiv2 doesn't provide a focal distance estimation for Sony yet. See [Exiv2/exiv2#582](https://github.com/Exiv2/exiv2/issues/582)

It's amazing how many headaches such a "simple" thing can involve, isn't it? So, then, we should hold off merging this in the meantime? Did you get a chance to check if the default infinity corrections do anything unpleasant under the various circumstances discussed?

I did try putting a large piece of white acrylic in a window of an otherwise dark room, hoping to find a distancedistance from the lens face where the vignetting became less over-emphasized, but the lens I was profiling didn't produce any meaningful results... I'll try with another lens when I get a chance... along with a couple of other "mad science" experiments I've been pondering.

The idea of just profiling at whatever focus distance produces the strongest vignetting keeps sounding better and better. ;)

cryptomilk:
Can this be merged?

junkyardsparkle:
I'll merge it, but can you harmonize the the model name format with the rest of the db first (basically just Fn.n > f/n.n)? Also, the Samyang change makes no sense, as far as I can tell. The model+lang tags don't function as aliases, they're just for adding a "pretty" name when needed... and case doesn't matter for matching anyway: https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/calibration.rst#create-lensfun-data

cryptomilk:
Updated, I didn't know about the pretty naming. I will add it to the howto on pixls.us.

junkyardsparkle:
The most common use for the localisation of names is for fixed-lens cameras, as seen in the "compact-*" db files.

PR #586, opened 2019-04-16
Sony: Add distortion for Sony E 18-135mm

scosu:
Signed-off-by: Markus Pargmann <mpa@devmp.org>

### DIFF
--- a/data/db/slr-sony.xml
+++ b/data/db/slr-sony.xml
@@ -989,4 +989,24 @@
         </calibration>
     </lens>
 
+    <lens>
+        <maker>Sony</maker>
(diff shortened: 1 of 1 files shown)


matt-maguire:
This is an E-mount lens, and E-mount lenses are for Sony mirrorless cameras, not Sony SLRs. Therefore, the correct place to put corrections for this lens is in mil-sony.xml rather than slr-sony.xml.

Looking at mil-sony.xml, it seems we already have a full set of corrections for this lens, including distortion, TCA and vignetting.  Therefore, I will close this PR. Thanks for your contribution!

PR #872, opened 2019-05-01
Appveyor: Add Appveyor MinGW-w64 and MSVC

1480c1:
Adds mingw-w64 testing via msys2

Adds MSVC testing. Does not work at the moment due to the GLIB2 not being found / not being attempted to be found


1480c1:
Going to close this temporarily until I can get cmake to work well with finding glib2 with vcpkg.

Issue #544, opened 2019-04-13
Calibration upload e34e56


bronger:
Calibration images were uploaded to the directory that starts with “`e34e56_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> SIGMA 16mm f/1.4 (c) AF DC DN E-mount

tgurr:
@bronger I'm afraid it looks like the "`e34e56_`" directory doesn't contain any images at all, should I start from scratch or use that directory/this GitHub issue anyways?

bronger:
@tgurr @lensfun/calibrators There is a bug in the calibration server so that the filenames are not properly escaped.  Then, ownCloud refuses to sync them. I fixed the problem for the current uploads, but future uploads may suffer the same problem until this is fixed.

tgurr:
@bronger I'm currently running into an issue with my first tests of the calibration data I produced, I use RawTherapee 5.6 as well as darktable 2.6.2 to have a look at the calibration. Both programs are able to auto-detect the required lensfun data, but they behave differently. In EXIF the lens is called `16mm F1.4 DC DN | Contemporary 017`, for darktable the auto-detect works with `<model>Sigma 16mm f/1.4 DC DN | Contemporary 017</model>` but for RawTherapee it needs to read `<model>Sigma 16mm f/1.4 DC DN | C</model>` so to have everthing working I need the following lensfun xml:

```
    <lens>
        <maker>Sigma</maker>
        <!-- RawTherapee model entry -->
        <model>Sigma 16mm f/1.4 DC DN | C</model>
        <mount>Sony E</mount>
        <cropfactor>1.534</cropfactor>
        <calibration>
            <distortion model="ptlens" focal="16" a="0.0198335" b="-0.0623523" c="0.0293053"/>
        </calibration>
    </lens>
    <lens> 
        <maker>Sigma</maker> 
        <!-- darktable model entry -->
        <model>Sigma 16mm f/1.4 DC DN | Contemporary 017</model>
        <mount>Sony E</mount>
        <cropfactor>1.534</cropfactor>
        <calibration>
            <distortion model="ptlens" focal="16" a="0.0198335" b="-0.0623523" c="0.0293053"/>
        </calibration>
    </lens>
```
[...]

PR #939, opened 2019-07-17
Add SIGMA 16mm F1.4 DC DN | C for Sony E from upload e34e56

tgurr:
Fixes <#544>.

### DIFF
--- a/data/db/mil-sigma.xml
+++ b/data/db/mil-sigma.xml
@@ -224,4 +224,16 @@
         </calibration>
     </lens>
 
+    <lens>
+        <maker>Sigma</maker>
+        <model>Sigma 16mm f/1.4 DC DN | Contemporary C 017</model>
+        <mount>Micro 4/3 System</mount>
+        <mount>Sony E</mount>
+        <cropfactor>1.534</cropfactor>
+        <calibration>
+            <!-- Taken with Sony 6000 (ILCE-6000) -->
+            <distortion model="ptlens" focal="16" a="0.019888" b="-0.0654977" c="0.0396784"/>
+        </calibration>
+    </lens>
+
 </lensdatabase>


tgurr:
Still would like to hear if there's some better approach than to have the entry two times for darktable and RawTherapee to properly auto-detect the right lensfun entry. Everything else I tried, like two model entries in one dataset failed :cry:.

bronger:
Two entries should be merged because Lensfun is supposed to be neutral with respect to the application program.  You may try the union of both names, i.e. something that contains all tokens.

Sigma 16mm f/1.4 DC DN | Contemporary C 017

tgurr:
@bronger thanks! That appears to work indeed for both of them, updated the PR.

data/db/mil-sigma.xml
```diff
@@ -224,4 +224,15 @@
         </calibration>
     </lens>
 
+    <lens>
+        <maker>Sigma</maker>
+        <model>Sigma 16mm f/1.4 DC DN | Contemporary C 017</model>
+        <mount>Sony E</mount>
```
bronger: Is this really the only mount?

tgurr: The lense is also available for Micro 4/3, but that one is already covered by https://github.com/lensfun/lensfun/commit/dc7557d5c8f5bc62e572ef1d462efdfcf87a5107 and has different calibration data as far as I can judge, also the results doesn't look good for my photos when choosing the already avalable Micro 4/3 entry manually. I'm new to this stuff so I might be missing how to properly combine data within the xml.

bronger: Ah, good to know. We do the following in such cases: Put the two datasets next to each other, add all mounts to both entries. In a future Lensfun version, we hope to be able to merge such duplicates.

tgurr: > Put the two datasets next to each other

check
> add all mounts to both entries

did that for the new entry, not sure if adding emount to the already existing should be part of this PR. I could do a follow-up PR or add the emount to the existing entry with a second commit in this PR. What do you prefer?

bronger: You’re right, I do that separately.

PR #950, opened 2019-07-30
Reformatted LEICA DG 8-18/F2.8-4.0 model name

tomalakgeretkal:
Doesn't match the conventional format. Causes some problems downstream. Hope this is okay.

(Also shouldn't these lenses be in mil-leica.xml?)

### DIFF
--- a/data/db/mil-panasonic.xml
+++ b/data/db/mil-panasonic.xml
@@ -1956,7 +1956,7 @@
 
     <lens>
         <maker>Panasonic</maker>
-        <model>LEICA DG 8-18/F2.8-4.0</model>
+        <model>Leica DG 8-18mm f/2.8-4.0</model>
         <model lang="en">Leica DG Vario-Elmarit 8-18mm f/2.8-4 Asph.</model>
         <mount>Micro 4/3 System</mount>
         <cropfactor>2</cropfactor>


veluria:
You cannot just rename the lens. It must match the exif data: 

    Exif.Photo.LensModel LEICA DG 8-18/F2.8-4.0

tomalakgeretkal:
That's unfortunate, because that string is "wrong" (doesn't match real-world materials) and ExifTool produces the "right" string via LensID.

Not all camera/lens combos provide a LensModel, so matching LensModel _can't_ be a general rule. LensID is already a much better general-purpose comparison, and since I'm using that, the current string fails lookup for me. :(

junkyardsparkle:
LensID is just a number, not an actual string (try exiftool -n). Software has to map it to a string, and it's not always straightforward: https://dev.exiv2.org/projects/exiv2/wiki/Lens_Recognition_in_Exiv2_v026_(and_later)/

AFAIK the goal is still to match exiv2, since that's what most of the software using lensfun is using.

As far as matching "real world materials" or just making things pretty, that's what this line is for:
`<model lang="en">Leica DG Vario-Elmarit 8-18mm f/2.8-4 Asph.</model>`

tomalakgeretkal:
I know what LensID is; I've contributed to it. It's a [generated] string. I believe you are referring to LensType (off the top of my head) which is indeed mapped. ExifTool uses various data sources, including LensModel and LensType, to generate a string that matches the actual, marketed name of the lens (most of the time!). It's this that I was hoping we'd strive for in LensFun, but if that's not the case then I'll just work around that.

junkyardsparkle:
```
$  exiftool -n '-lens*' 20191030-IMG_0064.CR2
Lens Type                       : 240
Lens Info                       : 17 55 0 0
Lens Model                      : EF-S17-55mm f/2.8 IS USM
Lens Serial Number              : 0000101e62
Lens                            : 17
Lens ID                         : 240
Lens                            : 26.4230351008965
```

junkyardsparkle:
> The "en" language variants are promising, but many across the DB do not follow this pattern. Perhaps that's where the potential improvement lies, then?

Yep, that would be the place to make changes for anything other than strictly matching purposes.

tomalakgeretkal:
@junkyardsparkle The Lens ID number you're looking at it merely an artefact of the software's architectural machinery.

Lens ID is a "composite field", which takes as input various potentially-present real metadata fields, and presents a string output as its display value. It has no inherent numeric value. When you jump into "raw" mode, you merely see the raw value of the first contributing "real metadata" field (as it happens, `LensType`), since there's nothing more useful to display. Ideally it would be blank but that's how it is at the moment.

You can study the ExifTool source to learn more about this. Lens ID generation is at Exif.pm:4648 and Exif.pm:5176. I'm more than happy to answer more questions about ExifTool as I'm quite familiar with its innards.

junkyardsparkle:
> Lens ID does not come from the file.

Thanks, I probably was confusing it with LensType. As far as the generated strings go, it might be nice if there was an independent project to maintain a "canonical" database, maybe even including some reliable heuristics for situations as described in that exiv2 link. The situation seems complex enough to warrant it... :-)

PR #968, opened 2019-08-26
Travis: Update macOS image

1480c1:
homebrew complains when not updated in a while, and since travis-ci does not seem to update their old images, either better upgrade the image or update hombrew (which takes a while)


seebk:
Thanks for the fixes.

Issue #973, opened 2019-09-03
Zeiss Batis 25mm f2 not recognized


vilim:
I am using lensfun through darktable, and there seems to be a bug in recognizing my lens (it is in the database, but is not being picked up automatically).

Exiv2 gives the following output:
```
Exif.Image.Make                              Ascii       5  SONY
Exif.Image.Model                             Ascii       9  ILCE-7M3
Exif.Photo.MakerNote                         Undefined 37496  (Binary value suppressed)
Exif.MakerNote.Offset                        Long        1  5310
Exif.MakerNote.ByteOrder                     Ascii       3  II
Exif.Sony2.SonyModelID                       Short       1  (363)
Exif.Sony2.LensID                            Long        1  Manual lens
Exif.Photo.LensSpecification                 Rational    4  250/10 250/10 20/10 20/10
Exif.Photo.LensModel                         Ascii      18  ZEISS Batis 2/25
Exif.Thumbnail.Make                          Ascii       5  SONY
Exif.Thumbnail.Model                         Ascii       9  ILCE-7M3
```

ZEISS Batis 2/25 is also written in Darktable. I would have added the alias to the xml via a pull request myself, but I am not sure of the procedure.

FreddieWitherden:
What does exiftool identity the lens as?

vilim:
`ZEISS Batis 2/25` as listed above 
Thanks!

FreddieWitherden:
So it looks as if the model name: ```        <model>E 25mm f/2</model>``` is wrong and this is why the lens is not being recognised.

vilim:
Exactly. But it is possible that other/older cameras output it as E 25mm f/2, which is why I guess it was put it like this in the first place. There is support for multiple lens aliases, right?

vilim:
> What does exiftool identity the lens as?

Sorry, I did not read this carefully, exiftool (unlike exiv2) lists 
```
Lens Type 3: Zeiss Batis 25mm F2
Lens Type 2: Zeiss Batis 25mm F2
Lens spec:﻿﻿ E 25mm F2
Lens Model: ZEISS Batis 2/25
```
I am not sure which of these fields is relevant.
I think darktable uses the Lens Model, which is also given by exiv2

tuxfanx:
completet

PR #983, opened 2019-09-09
Update so that Zeiss Batis 25mm f2 is recognized

vilim:
Resloves #973


FreddieWitherden:
Looks good to me.

vilim:
@uchrisu @bronger If it's ok, could this be merged please, so it gets integrated into the Darktable master?

junkyardsparkle:
For now I've added the former name back as a duplicate entry, due to reports of darktable still trying to match that name. Possibly depending on firmware updates?

Issue #1005, opened 2019-09-27
Calibration upload 2c4c76


bronger:
Calibration images were uploaded to the directory that starts with “`2c4c76_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Calibration photos made with Nikon Coolpix B700. Sadly, I had trouble finding an object which could be used as calibration picture at the high extreme of focal length (I would have to go about one hundred meters away which I couldn't due to trees everywhere...).

MR-DOS:
Hi, I'm the author of this set of photos. If there is anything wrong with the photos (I kind of expect so), please contact me either via GitHub or via e-mail p.polasek@sh.cvut.cz

junkyardsparkle:
After looking at these images, I think they might be ok. The image circle appears to not be concentric with the frame in the wide shots, but there may not be anything you can do about this (it's not unusual for superzooms to have some copy variation in this regard). I'll try some corrections, using center-shifting in Hugin if needed, and see what we get.

MR-DOS:
You can find the results from Hugin in closed pull request #1007. These were mostly unusable.
I tried taking another set of images with stabiliser turned off. This was a bit better, but for images with lowest possible focal length, I have to set something like 2x larger focal length in darktable in order to get reasonable results.
```
    <lens>
        <maker>Nikon</maker>
        <model>Coolpix B700</model>
        <mount>nikonB700</mount>
        <cropfactor>5.62</cropfactor>
        <calibration>
           <distortion model="ptlens" focal="4.3" a="-0.020983" b="-0.036141" c="-0.074442" />
           <distortion model="ptlens" focal="5.9" a="0.027473" b="-0.098339" c="-0.002399" />
           <distortion model="ptlens" focal="9.0" a="-0.028088" b="0.095659" c="-0.256602" />
           <distortion model="ptlens" focal="12.5" a="0.001534" b="-0.011987" c="-0.015353" />
           <distortion model="ptlens" focal="18.8" a="0.008651" b="-0.03247" c="0.019088" />
           <distortion model="ptlens" focal="32.2" a="-0.003742" b="0.017412" c="-0.039156" />
           <distortion model="ptlens" focal="53.7" a="-0.029092" b="0.11628" c="-0.14725" />
[...]

PR #1007, opened 2019-09-29
Added distortion and TCA calibration for Nikon Coolpix B700

MR-DOS:
Hi, these calibration data were generated using pictures from issue #1005

### DIFF
--- a/data/db/compact-nikon.xml
+++ b/data/db/compact-nikon.xml
@@ -737,5 +737,48 @@
             <tca model="poly3" focal="42.8" vr="0.9997493" vb="0.9998575"/>
         </calibration>
     </lens>
-
+    
+    <lens>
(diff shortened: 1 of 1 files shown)


MR-DOS:
Sadly, it seems the calibration causes enormous distortion on short focal lengths.

junkyardsparkle:
@MR-DOS For fixed-lens cameras, you need to also add a profile for the camera itself. You can see examples at the top of the `compact-nikon.xml` file. It looks like some of those close-together focal lengths could probably be left out, but it's hard to say which ones without looking at the pictures. I'll try to do that soon. One thing that might be important when taking calibration shots with this camera is to disable the image stabilization, if possible, since this can produce images where the distortion is off-center in the frame.

PR #1014, opened 2019-10-06
Catch unhandled exception in lensfun-update-data

deeno35:
lensfun-update-data is crashing on OSX due to an unhandled exception.  Fix

Also improve the error message to include the actual error


junkyardsparkle:
Thanks!

PR #1039, opened 2019-11-18
Another name for the Sigma 18-300mm

goddisignz:
Another name for the Sigma 18-300mm used on Sony Alpha 6300 with MC-11 adapter


xiota:
The Contemporary, Art, Sports lenses are different lenses, not just alternate names.

goddisignz:
Is there a non-Contemporary version of the 18-300mm lens? At least Sigma only lists the Contemporary.

Sturmflut:
The MC-11 converter seems to be for Sony E mount, but the lens entry that this commit is being applied on only has an entry for Sony Alpha (Sony A/Minolta) mount.

According to Sigma there is only a "Contemporary" version, yes.

Issue #502, opened 2019-03-19
Check if the C-interface is still up to date and fully functional


seebk:
Check if the C-interface is still up to date and fully functional.

Already known issues:

- Functions like lfModifier::EnableDistortionCorrection (const lfLens* lens, float focal) are available from C-interface, but lfModifier::EnableDistortionCorrection (const lfLensCalibDistortion& lcd) are missing

letmaik:
- `lf_db_load_data`: I opened https://github.com/lensfun/lensfun/pull/1040 for that.
- `lf_db_read_timestamp`: There is no issue, the function is properly defined and exported.
- `lf_db_save`: The signature changed: definition has `char *lf_db_save (const lfDatabase *db)`, declaration has `char *lf_db_save (const lfMount *const *mounts, const lfCamera *const *cameras, const lfLens *const *lenses)`. I would break the old API and simply rename it to `char *lf_db_save_data (const lfDatabase *db)` to be consistent with `lf_db_load_data`.
- `lf_db_save_file`: This function is declared as `lfError lf_db_save_file (const lfDatabase *db, const char *filename, const lfMount *const *mounts, const lfCamera *const *cameras, const lfLens *const *lenses)` but defined as `lfError lf_db_save_file (const lfDatabase *db, const char *filename)`. I suggest breaking the old API and using the definition signature, together with deprecating `lfError lf_db_save_all (const lfDatabase *db, const char *filename)` which does the same.
- `lf_mount_copy`: Nothing actually changed between 0.3.2 and master. The function is declared but not defined. I suggest removing the declaration from the header. No one could have used it anyway.

Please confirm whether the above is fine and I can start submitting the remaining PRs.

seebk:
The second item is now fixed in master.

PR #1040, opened 2019-11-24
Fix lf_db_load_data

letmaik:
`lf_db_load_data` had the wrong signature in the implementation not matching the header. This PR fixes that. Also, `Load (const char *errcontext, const char *data, size_t data_size)` is made public again (so that it can be called from `lf_db_load_data`) but marked as `DEPRECATED`. Note the docstring was copied from the last release (minus the sentence about deprecation).

This fixes an issue from #502. Re-designing the load functions is not part of this PR, but simply to get lensfun into a working state again.

### DIFF
--- a/include/lensfun/lensfun.h.in
+++ b/include/lensfun/lensfun.h.in
@@ -1823,6 +1823,25 @@ struct LF_EXPORT lfDatabase
     lfError Load (const char *data, size_t data_size);
 
     /**
+     * @brief Load a set of camera/lenses from a memory array.
+     *
+     * This function is deprecated and will be removed in the future.
+     * Please use the lfDatabase::Load(const char *data, size_t data_size) syntax instead.
+     * 
+     * This is the lowest-level loading function.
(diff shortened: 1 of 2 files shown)


seebk:
I have committed a slightly different fix and simply changed the signature in the header to match the implementation. Hope that resolves your issue, too. The errcontext IMO is an implementation detail that should not bother the regular programmer using Lensfun and I would prefer to hide this parameter.

seebk:
Or is it necessary to mark it deprecated? I am not sure if anyone ever used 'lf_db_load_data' before...

letmaik:
I did, but looks like I'm the only one, at least according to a code search on github. Your patch is fine, no need to keep the old one and deprecate it. I'll deal with it in my own code accordingly.

seebk:
Ok, I marked it as deprecated now. `lf_db_load_data` actually is not a good name because it reads from a string char array. So we want to rename it anyway...

PR #1044, opened 2019-12-04
Added camera entry for Canon EOS Rebel T6s

EdgarLux:
Added an alternative name for Canon EOS 760D  model.

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -473,6 +473,14 @@
         <mount>Canon EF-S</mount>
         <cropfactor>1.613</cropfactor>
     </camera>
+	
+    <camera>
+        <maker>Canon</maker>
+        <model>Canon EOS Rebel T6s</model>
+        <model lang="en">EOS Rebel T6s</model>
+        <mount>Canon EF-S</mount>
+        <cropfactor>1.613</cropfactor>
+    </camera>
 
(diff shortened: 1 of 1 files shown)


gnl21:
LGTM. Do you want to review changes like this, @bronger, or shall I merge it?

junkyardsparkle:
@gnl21 There's no real harm in adding a camera, it's easy enough to change if necessary. You might want to add a comment mentioning it's an alias for 760D, in case somebody wants to convert those entries into aliases some day.

gnl21:
Sure, thanks, I just don't want to step on anyone's toes.

I think the comment is a good idea (and putting them in the same entry one day would be even better), but basically all of these rebel cameras have multiple names and none of them have comments. I'll merge this PR now and we can add comments to all of these cameras together.

Thanks for the contribution, @EdgarLux.

junkyardsparkle:
@gnl21 I guess each brand ecosystem is probably best dealt with by somebody who's familiar with it... I've actually become more familiar with Panasonic in this regard just from trying to make sense of their fixed-lens camera regional aliases (because those would involve redundant lens profiling otherwise).

EdgarLux:
> Thanks for the contribution, @EdgarLux.

Thanks to you. I'm pleased to have contributed a little.

PR #1086, opened 2020-02-07
Improve profiles for Sony 16-35mm f/2.8 and 85mm f/1.4

cryptomilk:
I've created better vignetting pictures and fixed an issue with TCA overcorrecting.


cryptomilk:
I've updated https://pixls.us/articles/create-lens-calibration-data-for-lensfun/ for creating better vignetting pictures so we do not brighten the picture.

PR #1093, opened 2020-02-14
Added Canon EOS KISS M camera

matt-maguire:
It's actually the same as the Canon EOS M50, just a version of it released for the Japanese market.

Added an entry to the camera database so darktable can automatically pick up the right lens
correction profile from the EXIF data.


junkyardsparkle:
Thanks!

PR #1118, opened 2020-03-10
Fixed name of Nikon AF-S DX Zoom-Nikkor 18-55mm f/3.5-5.6G ED

blastrock:
Hi,

I have this exact lens, and while it seems to be supported by lensfun, its name does not match the one extracted by exiv2.

Here is the relevant output from exiv2 0.27.2:

```
Exif.NikonLd3.LensIDNumber                   Byte        1  Nikon AF-P DX Nikkor 10-20mm f/4.5-5.6G VR
```

I am not sure this is the right fix, maybe we should support both names?


blastrock:
Rebased on master. Bump?

Sturmflut:
There is some confusion here.

Lens in the title: AF-S DX Zoom-Nikkor 18-55mm f/3.5-5.6G ED

Lens in the exiv2 output and in the commit: Nikon AF-P DX Nikkor 10-20mm f/4.5-5.6G VR

Those are two different lenses. I was able to find some test shots though, and exiv2 actually identifies the lens as "Nikon AF-P DX Nikkor 10-20mm f/4.5-5.6G VR" (as one would expect). Since the branch now has conflicts, I simply added another <model> tag (you never know who is using the old name) and commited the change to master.

Fixed in lensfun@437817ab93cd8c4ef67aa790d204950b5cdcb6fb

Sturmflut:
There is some confusion here.

Lens in the title: AF-S DX Zoom-Nikkor 18-55mm f/3.5-5.6G ED

Lens in the exiv2 output and in the commit: Nikon AF-P DX Nikkor 10-20mm f/4.5-5.6G VR

Those are two different lenses. I was able to find some test shots though, and exiv2 actually identifies the lens as "Nikon AF-P DX Nikkor 10-20mm f/4.5-5.6G VR" (as one would expect). Since the branch now has conflicts, I simply added another <model> tag (you never know who is using the old name) and commited the change to master.

Fixed in lensfun@437817ab93cd8c4ef67aa790d204950b5cdcb6fb

blastrock:
Sorry, I messed up the commit message indeed. I didn't know we could put multiple `model` tags, thank you for taking care of this!

PR #1129, opened 2020-03-19
Remove shebang from __init__.py.in

paolostivanin:
Usually, an `__init__` file doesn't need a shebang or, if it does, it should be marked as executable.


bronger:
Please leave it. I am accustomed to recognise Python files by that line. Besides, it doesn’t do any harm.

bronger:
I gave it more thought and came to the conclusion that it is slightly better to have the shebang only for executable scripts. I can adapt.  :-)

Issue #1103, opened 2020-02-27
TAMRON 28-75 f/2.8 Di III RxD for SONY eMount is not in lensfun db


peter-hb-1955:
Hello,
I missed my TAMRON 28-75 f/2.8 DI RxD for SONY eMount in the lensfun database and therefore generated it by myself. Please find my data attached. I tested it with darktable and I think it is Ok to distribute for all.
All the best
[lensfun_calibration.zip](https://github.com/lensfun/lensfun/files/4262862/lensfun_calibration.zip)

Peter Seifert

StefanDieke:
Thanks a lot Peter, for making your calibration data for the Tamron DI RxD lenses for Sony eMount available! This is exactly what I was looking for! I hope they will be integrated into lensfun quickly!
How do I integrate them locally (I'm running on linux)?

peter-hb-1955:
Hello Stefan,
you have to edit the lensfun file "mil-tamron.xml" and to add there the new <lens> data block before the file end </lensdatabase>. In manjaro KDE I have the kate editor. In other Linux versions there are different editors. The file has to be edited / saved in admin mode. For Manjaro KDE YOu can edit it in the terminal with this command "kate /usr/share/lensfun/version_2/mil-tamron.xml". May be in other linux the lensfun directory is different.
In my case I also had to add my camera in the file "mil-sony.xml". The ALPHA 7 RM3 data can be duplicated simply from the R2 data block and then changing the camera name.  Darktable is recognizing the lens only if also the camera is available in lensfun database.
Have fun!
Peter Seifert

StefanDieke:
Perfect! Thanks a lot again, Peter! After finding the right files (which for Antergos are in ~/.local/share/lensfun/updates/version2) it works like a charm!
Anyway, I think this data should be included in the official lensfun database for a wider usage!

vladakoz:
Hello, I am very new to this but I was not successfull following your process. I have Sony A 7 III body and Tamron 28-75mm f/2.8 Di III RXD lens. I am running Darktable on Linux Leap 15.1.
After implementing your procedure Darktable is not able to find the lens. In fact if I attempt to find the lens manually a do not have the Tamron available in the lens pull down menu. The camera is reported as Sony, Alpha 7 III and the lens as E 28-75mm F2.8-2.8. Any suggestions where I went wrong ?
Vlada.

uchrisu:
[...]

PR #1133, opened 2020-04-04
Update mil-sigma.xml

uchrisu:
Added data for the Tamron 28-75 2.8 for Sony E as provided by @peter-hb-1955 in issue  #1103.
Tested the profile on own pictures. Results are convincing.


uchrisu:
Updated wrong file

PR #1151, opened 2020-04-24
Give User-Profiles always priority

### LINKED ISSUE #593: lensfun-update-data breaks added adapters
Using lensfun 0.3.95.

I had added adapters using lensfun-add-adapter and all is ok. Ran lensfun-update-data, update successfully performed and new data in /home/myusername/.local/share/lensfun.

lensfun-update-data Creates a directory "updates/version_2" with the new data (works). But (at least using Darktable) the /usr/local/share/lensfun/version_2/_mounts.xml file is ignored so adapted lenses are no longer available.

Running lensfun-add-adapter again and adding the adapters again creates a file /home/myusername/.local/share/lensfun/_mounts.xml. This is apparently in the wrong location since moving the file to /home/myusername/.local/share/lensfun/updates/version_2 fixes the problem.

Long story short, either do not ignore the standard file /usr/local/share/lensfun/version_2/_mounts.xml or have users rerun lensfun-add-adapter but then the _mounts.xml has to be put in the proper location by lensun-add-adapter.

uchrisu:
Profiles added by the users should always have priority.
This also fixes  #593 which is marked as milestone for stable release.

### DIFF
--- a/libs/lensfun/database.cpp
+++ b/libs/lensfun/database.cpp
@@ -100,6 +100,9 @@ lfError lfDatabase::Load ()
         ReadTimestamp (SystemUpdatesLocation);
     const int timestamp_user_updates =
         ReadTimestamp (UserUpdatesDir);
+    
+    Load (HomeDataDir);
+    
     if (timestamp_system > timestamp_system_updates)
         if (timestamp_user_updates > timestamp_system)
             err = Load (UserUpdatesDir);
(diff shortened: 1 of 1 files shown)


seebk:
According to the docs, profiles from database locations loaded later will override previously loaded profiles. If that's true the previous implementation was correct. User profiles loaded with `Load (HomeDataDir);` should always override any update or system profiles. Or maybe something else in the folder loading is off here. Anyway I fear this proposed fix is not sufficient.

Without really debugging it I guess the problem is that the prefix for `HomeDataDir` and `UserUpdatesDir` is the same and `Load()` recurses into subfolders.  Calling `Load (HomeDataDir);` will do the following loads:

/home/myusername/.local/share/lensfun/_mounts.xml
/home/myusername/.local/share/lensfun/updates/*.xml
...

Hence, the updates subfolder content will again override the previous user mounts.

On first glance I'd say the `/home/myusername/.local/share/lensfun/updates location` has to move, maybe to `/home/myusername/.local/share/lensfun/updates`.

PR #1154, opened 2020-04-25
mil-sigma.xml

andrewlumley:
Fixed Lens names:

Sigma 16mm f/1.4 DC DN | Contemporary C 017 -> Sigma 16mm f/1.4 DC DN | C 017
"Contemporary" is redundant, since the "C" in the name stands for contemporary

SIGMA 30mm f/1.4 DC DN | C 016 -> Sigma 30mm f/1.4 DC DN | C 016
Fixed capitalization of "Sigma" to match other Sigma lenses

### DIFF
--- a/data/db/mil-sigma.xml
+++ b/data/db/mil-sigma.xml
@@ -184,7 +184,7 @@
 
     <lens>
         <maker>Sigma</maker>
-        <model>SIGMA 30mm f/1.4 DC DN | C 016</model>
+        <model>Sigma 30mm f/1.4 DC DN | C 016</model>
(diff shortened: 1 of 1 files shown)


uchrisu:
Hi, we can not simply change names (i.e., "model") since this has to coincide with exif information. For a nice "printing" name there are the language options, e.g., <model lang="en">Sigma 16mm f/1.4 DC DN C</model>

When we would change the "model" name, the lens can not be automatically detected for corrections anymore. So a program such as darktable would not automatically choose the correction data anymore and the lens has to be manually chosen.

tuxfanx:
nothing changed up to now --> closed

Issue #1155, opened 2020-04-25
Calibration upload 66ca80


bronger:
Calibration images were uploaded to the directory that starts with “`66ca80_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Sony E 70-350mm F/4.5-6.3 G OSS (SEL70350G)
> 8 calibration shots all taken at F/8.0: 70mm, 98mm, 132mm, 163mm, 200mm, 244mm, 291mm, 350mm.

andrewlumley:
I am the uploader. All photos taken with my Sony a6600 at F/8.0 and ISO 100.

uchrisu:
Looks like a great job again! Thank you. I think I will do this tomorrow.

andrewlumley:
No, thank you! The last profile you made has been working great. Thanks for doing this!

uchrisu:
So, it was hard work, since there was only one trustful horizontal line and an "a-only" fitting did not give useful results. However, also fitting to vertical lines delivered a really good result. I checked it with darktable and drawing straight lines in gimp afterwards. 

I will add it to master. 

`<lensdatabase>

    <lens>
        <maker>Sony</maker>
        <model>E 70-350mm F4.5-6.3 G OSS</model>
        <mount>Sony E</mount>
        <cropfactor>1.534</cropfactor>
        <calibration>
            <distortion model="ptlens" focal="70" a="0.0102674203235291" b="-0.0226289059168314" c="0.052844698199936"/>
            <distortion model="ptlens" focal="98" a="0.00651089549890782" b="-0.00339194289178117" c="0.0266660460095801"/>
            <distortion model="ptlens" focal="132" a="0.00120680770839538" b="0.0103273006109119" c="0.020987478836464"/>
            <distortion model="ptlens" focal="163" a="-0.0042520752612236" b="0.0297037217037091" c="-0.00242810977838033"/>
            <distortion model="ptlens" focal="200" a="-0.00219146298780921" b="0.0315922868063698" c="-0.0266377348354084"/>
            <distortion model="ptlens" focal="244" a="-0.0162024853968891" b="0.0803244662732511" c="-0.086079101637817"/>
            <distortion model="ptlens" focal="291" a="-0.0107445491144207" b="0.0680949523917039" c="-0.0895930575079237"/>
[...]

PR #1156, opened 2020-04-26
Added Sony E 70-350mm F4.5-6.3 G OSS

uchrisu:
distortion and tca, thanks to @andrewlumley for the pictured deliver in #1155


andrewlumley:
Thanks again!

PR #1166, opened 2020-04-27
Make plots more readable

uchrisu:
Especially for slightly color blinded people the "fit" line was not visible before. Not it should be more easy to check the results.


uchrisu:
@bronger Since it is your script I ask you for reviewing and merging. 
I had really problems to see the line before, while the "fit" line is very important. With the change everything is fine now.

Issue #945, opened 2019-07-24
Scaling factor API changes break existing image edits


jakubfi:
Scale factor used by lfModifier::EnableScaling() and lfModifier::GetAutoScale() has different meaing in 0.3.2 and git lensfun versions. This results in broken image edits done in a program compiled with stable lensfun version when opened in a program compiled with current development lensfun version.

Actual case: When lens correction is applied in darktable compiled against lensfun-0.3.2 sidecar file is written with, say, scaling factor set to 1.033 (as computed by the library). Opening the same edit in darktable compiled against lensfun-git sets the scaling factor to 1.033 again, causing distortion correction artifacts on the edge of the image. When lens corrections are re-applied, scaling factor is now set to 1.047 and image looks good again.

Automatically resetting the scale factor in darktable after switching to lensfun-next from 0.3.2 is not a solution, because user may set have it to a deifferent, preferred value (which API allows for).

lensfun version should not redefine the actual meaning of the scaling factor (even if it was wrong in the beginning). Instead, it should do all the necessary back and forth calculations behind the scenes of EnableScaling() and GetAutoScale(), and keep the scaling API consistent across versions.

seebk:
Not sure if we can convert the scale factor or if it makes sense. Maybe @bronger has an idea?

I will put it to the next release milestone.

jakubfi:
@seebk @bronger If converting the scale factor behind the scenes is not feasible, maybe provide a method to allow external applications to convert it when necessary? I don't know all the lensfun usage scenarios, but I would assume that darktable is not the only one to bump into this in the future.

uchrisu:
Okay, it seems to be due to the "ru" stretching as distortion parameters are converted to keep the real focal length. "ru" is internally stretched by factor "d". I would suggest to add member d in lfLensCalibDistortion wehn transforming the distortion coefficients. Then GetAutoScale and EnableScaling can use this "d" factor by simply dividing/multiplying the d factor to get back to the former scaling factor. 

I will implement this and check the results against the results from 0.3.92 if no one complains about this way.

uchrisu:
However, changing back the meaning of the scale we loose the convention that scale=1 preserves the real focal length. 

Another problem: Many people are already using 0.3.95 at the moment. When we bring back the scale interpretation as in 0.3.92 pictures get broken for them. 

Maybe we should keep it as it is now? Or should we add at least functions like GetREALAutoScale and EnableREALScaling (without all Caps, of course)?

[...]

PR #1174, opened 2020-05-09
Scaling API backward compatibility to 0.3.2

uchrisu:
I changes the scaling factor in the API back to the meaning of 0.3.2 release as demanded by issue #945 .

I exemplarily verified it for a fake lens with very high distortion:
[test_scaling.zip](https://github.com/lensfun/lensfun/files/4604203/test_scaling.zip)

Still, there are very small differences left over (about 1 pixel) but they seem to come from further changes, e.g., the center of the image has changed since 0.3.2 a tiny bit.

### DIFF
--- a/libs/lensfun/mod-coord.cpp
+++ b/libs/lensfun/mod-coord.cpp
@@ -34,7 +34,7 @@
 #include <limits>
 
 lfLensCalibDistortion rescale_polynomial_coefficients (const lfLensCalibDistortion& lcd_,
-                                                       double real_focal, cbool Reverse)
+                                                       double real_focal, cbool Reverse, double *scaling_d)
 {
     // FixMe: The ACM probably bases on the nominal focal length.  This needs
     // to be found out.  It this is true, we have to scale its coefficient by
@@ -43,11 +43,12 @@ lfLensCalibDistortion rescale_polynomial_coefficients (const lfLensCalibDistorti
     const float hugin_scale_in_millimeters =
         hypot (36.0, 24.0) / lcd.CalibAttr.CropFactor / hypot (lcd.CalibAttr.AspectRatio, 1) / 2.0;
     const float hugin_scaling = real_focal / hugin_scale_in_millimeters;
+    float d = 1;
     switch (lcd.Model)
     {
         case LF_DIST_MODEL_POLY3:
         {
-            const float d = 1 - lcd.Terms [0];
+            d = 1 - lcd.Terms [0];
             lcd.Terms [0] *= pow (hugin_scaling, 2) / pow (d, 3);
             break;
         }
@@ -57,7 +58,7 @@ lfLensCalibDistortion rescale_polynomial_coefficients (const lfLensCalibDistorti
             break;
         case LF_DIST_MODEL_PTLENS:
         {
-            const float d = 1 - lcd.Terms [0] - lcd.Terms [1] - lcd.Terms [2];
+            d = 1 - lcd.Terms [0] - lcd.Terms [1] - lcd.Terms [2];
             lcd.Terms [0] *= pow (hugin_scaling, 3) / pow (d, 4);
             lcd.Terms [1] *= pow (hugin_scaling, 2) / pow (d, 3);
             lcd.Terms [2] *= hugin_scaling / pow (d, 2);
@@ -67,12 +68,14 @@ lfLensCalibDistortion rescale_polynomial_coefficients (const lfLensCalibDistorti
             // keep gcc 4.4+ happy
             break;
     }
+    // This is needed for scaling compatibility to 0.3.2
+    *scaling_d = d;
     return lcd;
 }
 
 int lfModifier::EnableDistortionCorrection (const lfLensCalibDistortion& lcd_)
 {
-    const lfLensCalibDistortion lcd = rescale_polynomial_coefficients (lcd_, RealFocal, Reverse);
+    const lfLensCalibDistortion lcd = rescale_polynomial_coefficients (lcd_, RealFocal, Reverse, &scaling_d);
     if (Reverse)
         switch (lcd.Model)
         {
@@ -459,6 +462,8 @@ float lfModifier::GetAutoScale (bool reverse)
     // that we really have no black borders left.
     scale *= 1.001;
     scale *= subpixel_scale;
+    
+    scale *= scaling_d;
 
     return reverse ? 1.0 / scale : scale;
 }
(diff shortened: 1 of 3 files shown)


seebk:
Thanks a lot @uchrisu for creating that PR. 

As far as I can see it might be better to not scale the `scale` value in `GetAutoScale()` and `EnableScaling()` but to only incorporate the `scaling_d` value in `ModifyCoord_Scale()`. Otherwise there might be a problem when `EnableScaling()` is called before `EnableDistortion()` and the `scaling_d` value has not yet been set. Or did I miss something?

uchrisu:
@seebk as I understood in issue #945 the problem is the change of the meaning of the `scale` value in `GetAutoScale() ` and `EnableScaling()` . E.g., Darktable has saved the old values for edited images and when using a new lensfun version the meaning of the parameters you get from `GetAutoScale() ` and deliver to `EnableScaling()` have to be the same. Therefore, the only way is to change the meaning of the `scale` parameters back again as far as I understood the problem. A pure internal change will probably not help here.

Furthermore the documentation for `GetAutoScale` says
> This expects that all coordinate distortion callbacks are already set up and working.

So probably we are quite save to preserve the API meaning.

Maybe I understood it wrong. 
I can backport the changes to a local 0.3.95-alpha and test the behavior in Darktable. Probably this will show if everything works as intended.

seebk:
I see, you are right. In `GetAutoScale` the `scale` value is multiplied by `scaling_d` and in `EnableScaling` it is divided to get back to the internal meaning of the value. As `GetAutoScale` is only allowed to called after all callbacks are set up this is fine.

But we should still consider moving the inverse scaling in `EnableScaling()` to the `ModifyCoord_Scale()` callback. The following is absolutely valid from the API and otherwise could be a problem because `scaling_d` is not yet set.

    mod->EnableScaling(3.0);
    mod->EnableDistortionCorrection();

I'm on the go and can't fully see the code so I might  be wrong :) Just want to be sure we consider all cases...

uchrisu:
scaling_d is initialized to 1 in the constructor of `lfModifier`. 

However, your thought raise a completely new question: What should happen if `EnableScaling` is not used at all?
Since the distortion correction is changed to keep the real focal length the image will be enlarged or shrunk compared to the 0.3.2 release. Should this also be compensated internally? If this is the case then a fix might become a little more complicated but probably still manageable.

libs/lensfun/mod-coord.cpp
```diff
@@ -34,7 +34,7 @@
 #include <limits>
 
 lfLensCalibDistortion rescale_polynomial_coefficients (const lfLensCalibDistortion& lcd_,
-                                                       double real_focal, cbool Reverse)
+                                                       double real_focal, cbool Reverse, double *scaling_d)
```
bronger, line 37: Why not call-by-reference here?

libs/lensfun/mod-coord.cpp
```diff
@@ -43,11 +43,12 @@ lfLensCalibDistortion rescale_polynomial_coefficients (const lfLensCalibDistorti
     const float hugin_scale_in_millimeters =
         hypot (36.0, 24.0) / lcd.CalibAttr.CropFactor / hypot (lcd.CalibAttr.AspectRatio, 1) / 2.0;
     const float hugin_scaling = real_focal / hugin_scale_in_millimeters;
+    float d = 1;
```
bronger, line 46: I’d prefer setting `scaling_d` in every case and leave `d` a const. After all, `d` has a special meaning, and is not present in the other models.

bronger:
I am with you that we should do something about it.  However, I recommend not to carry hacks due to old errors to the future. The algorithms are hard enough already. So, my goal would be to encapsulate this necessary ugliness as much as possible.

I know this has been suggested before but … why not adding a scale factor conversion function, or adding `EnableLegacyScaling`? Or a `legacy` bool parameter to `EnableScaling`

Moreover, I’d like to preserve the expectation that “scale by 2” changes the effective focal length by 2 as well. This is not the case now, is it?

uchrisu:
I would vote for the suggestions of @bronger since I think all changes were done with the intention of clear math/geometry in the end. Changing back the meaning of the scaling factor would foil this intention. 
However, I did propose this kind of fix because of the discussion in  #945 but I wouldn't object to withdrawing the proposal. 

In my opinion the handling of a new lensfun version would need a special treatment anyway due to the API changes and, thus, this extra treatment of the scaling factor can also be included.

bronger:
What I don’t know yet: Do we have all variables where we need them? This is frequently a problem when enhancing or changing Lensfun. (@uchrisu saw this himself when introducing `d_scaling`.) In particular, a conversion function that returns the correction factor between old and new way would need a *lot* of parameters. On the other hand, if its a method of lfModifier …

I propose (and only propose) to do it with a `legacy` flag for `EnableScaling` as plan A, and a conversion function as plan B.

uchrisu:
I agree with the suggestion of @bronger .
When we default the legacy to true it would also be compatible with 0.3.2, e.g., by

`GetAutoScale ( bool reverse, bool legacy=true ) `

however, probably this might not be the best way since the new scaling should be the new standard way we go!?

bronger:
This won’t be an issue since all programs have to be changed anyway to adopt the new API.

seebk:
Adding a legacy flag to the methods that need it and default it to false sounds good. Please note that there also is a scale parameter in the deprecated `lfModifier::Initialize` method that has to be updated. And be aware of the case when `EnableScaling` is called before `EnableDistortion`.

bronger:
The ordering of the calls should not matter. Other than that, I fully agree.

seebk:
So what I wanted to say is that the Enable... calls can appear in any order in the calling source code and the scale conversion has to properly handle that cases. Calling EnableDistortion after EnableScaling with the original PR did not convert the scale factors.

libs/lensfun/mod-coord.cpp
seebk: Yes, call by reference would be better

libs/lensfun/modifier.cpp
```diff
@@ -162,6 +163,9 @@ int lfModifier::EnableScaling (float scale)
         if (scale == 0.0)
             return false;
     }
+    
+    // backward compatibility to 0.3.2
+    scale /= scaling_d;
```
seebk, line 168: Watch out that `scaling_d` is not yet set to the correct value when `EnableScaling` is called before `EnableDistortion`

Issue #1213, opened 2020-07-18
ASAN error calling lensfun 0.3.95-2 from darktable (master)


codingdave:
Running darktable master (darktable 3.1.0+2475~g08eba104b) on manjaro linux I get this ASAN report on startup (every time).

If you need any more information, I am happy to provide that. @bronger you can also reach me in IRC #darktable with the nick saulus.

Running latest lensfun master does not compile.

Cheers
Dave

=================================================================
==180533==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000c87113 at pc 0x7fefb6fb91ca bp 0x7ffdd13c7e20 sp 0x7ffdd13c75c8
WRITE of size 4 at 0x602000c87113 thread T0
    #0 0x7fefb6fb91c9 in __interceptor_memcpy /build/gcc/src/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:806
    #1 0x7fef992a7f24 in lfMount::AddCompat(char const*) (/usr/lib/liblensfun.so.2+0x35f24)
    #2 0x7fef99289065  (/usr/lib/liblensfun.so.2+0x17065)
    #3 0x7fefb5ef1369 in g_markup_parse_context_parse (/usr/lib/libglib-2.0.so.0+0x55369)
    #4 0x7fef9928b102 in lfDatabase::Load(char const*, char const*, unsigned long) (/usr/lib/liblensfun.so.2+0x19102)
    #5 0x7fef9928cdb5 in lfDatabase::Load(char const*) (/usr/lib/liblensfun.so.2+0x1adb5)
    #6 0x7fef9928cdef in lfDatabase::Load(char const*) (/usr/lib/liblensfun.so.2+0x1adef)
    #7 0x7fef9928cc0c in lfDatabase::Load() (/usr/lib/liblensfun.so.2+0x1ac0c)
    #8 0x7fef993230b3 in init_global /home/david/workspace/darktable.git/src/iop/lens.cc:1256
    #9 0x7fefb68f3ff2 in dt_iop_load_module_so /home/david/workspace/darktable.git/src/develop/imageop.c:436
    #10 0x7fefb680b6b0 in dt_module_load_modules /home/david/workspace/darktable.git/src/common/module.c:49
    #11 0x7fefb68f5ee5 in dt_iop_load_modules_so /home/david/workspace/darktable.git/src/develop/imageop.c:1544
[...]

PR #1216, opened 2020-07-20
fixes heap-buffer-overflow, #1213

codingdave:
Should fix https://github.com/lensfun/lensfun/issues/1213


codingdave:
I have re-tested that PR and can confirm again, that it fixes the bug. Also the change is very easy to understand and it very obviously fixes an error.

codingdave:
Seems being duplicate of de954c952929316ea2ad0f6f1e336d9d8164ace0

PR #1222, opened 2020-07-27
updated Sigma 35mm f/1.4 DG HSM (A12)

MStraeten:
new measure for near and infinity focus

### DIFF
--- a/data/db/slr-sigma.xml
+++ b/data/db/slr-sigma.xml
@@ -2055,23 +2055,23 @@
             <!-- Taken with Pentax K-1 -->
             <distortion model="poly3" focal="35" k1="-0.00111476"/>
             <tca model="poly3" focal="35" vr="1.0001811" vb="0.9999722"/>
-            <!-- Taken with Canon 6D -->
-            <vignetting model="pa" focal="35" aperture="1.4" distance="10" k1="-2.1367" k2="2.3253" k3="-1.0155"/>
-            <vignetting model="pa" focal="35" aperture="1.4" distance="1000" k1="-2.1367" k2="2.3253" k3="-1.0155"/>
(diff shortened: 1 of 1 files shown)


matt-maguire:
Was there any issue with the existing measurements?

MStraeten:
I found quite different behavior with near and far focussing distances at least with my lens - so the correction for the far distance was too much on near distances resulting in edges being brighter than the center. So I did a measure for min and max focussing distance - so at least it’s possible to adjust the correction by changing the focal length parameter (at least with darktable)
But Maybe there’re differences between lenses ...

matt-maguire:
Ok, thanks for the update.

junkyardsparkle:
> But Maybe there’re differences between lenses ...

This is very normal, and is probably an issue with any profile which only provides vignetting corrections at infinity (which includes many of the profiles in the database)... unfortunately.

davidlbit:
@MStraeten Has this lens been removed from the Lensfun database? I just updated the database, and also checked the supported lenses on the website, but seems like I couldn't find it.

MStraeten:
it’s there in slr-sigma.xml   <model>Sigma 35mm f/1.4 DG HSM</model>
don’t rely too much those supported lenses lists - they’re often not properly maintained. Check the db instead - if it’s there, it’s not removed ;)
If you don’t see it in darktable then have a look at https://github.com/darktable-org/darktable/wiki/User's-portal#my-lens-is-wrongly-identified

davidlbit:
Okay, I checked again and found out that the lens is not available for any of the sony alpha cameras (I'm owning a Sony A7R iii myself). But I can see that it's in fact available for the canon mirrorless lineup. I assume that means that I need to follow the tutorial to generate a calibration xml file and then send it in via the mailing list?

MStraeten:
if you’re using the canon version of the lens via adapter the definition should match. Otherwise you need to do the measurement for your version

PR #1237, opened 2020-08-25
Add calibration data for Nikon AF-S 200-500 5.6E VR ED lens

Sturmflut:


matt-maguire:
Your commit contains `<lensdatabase> </lensdatabase>` tags, which is causing the CI tests to fail. Could you remove them and push an updated commit? thanks.

Sturmflut:
I noticed that distortion and vignetting can still be improved, so I decided to revoke this pull request and send a new one when I have the data.

PR #1260, opened 2020-10-04
Added support for Canon 5DS and 5DSR

criadoperez:


uchrisu:
I downloaded an Raw file for the DS-R and the exiv2-tool tells me the name is "Canon EOS 5DS R" with a space before the R. 
Can you check it on your side?

criadoperez:
Yes, I confirm it is with the space "Canon EOS 5DS R"

PR #1270, opened 2020-10-15
Uchrisu patch nikkor24 1.8

uchrisu:
TCA and Distortion thanks to #1178

The name detected by calibrate.py  and darktable did not coincide. I took the one of darktable now.
Both versions were somehow different from the conventional Nikon lens naming. 

I would suggest to take the darktable-compatible name


uchrisu:
Adding a lang=en model tag. Most other lenses of this kind do this.

Issue #1173, opened 2020-05-09
Calibration upload 3c5d92


bronger:
Calibration images were uploaded to the directory that starts with “`3c5d92_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Calibration images for the Canon EF 80-200mm f/4.5-5.6 II

privong:
I am the uploader of these calibration images. Please let me know if I can provide any additional information.

uchrisu:
@privong  Calibration images worked very well! After the branch has passed all tests I will merge it into master.

privong:
@uchrisu Great. Thanks for doing the calibration!

uchrisu:
It it now in the master database

PR #1272, opened 2020-10-15
Uchrisu patch canon 80 200 45 56

uchrisu:
Distortion and TCA from Upload #1173


John55h:
exiv2 and exiftool identify "Canon EF 80-200mm f/4.5-5.6" and "Canon EF 80-200mm f/4.5-5.6 II" as "Canon EF 80-200mm f/4.5-5.6" so the wrong calibration is applied to Canon EF 80-200mm f/4.5-5.6"
See:
https://github.com/Exiv2/exiv2/issues/1906

PR #1280, opened 2020-10-22
Added different naming for same lens Canon EF 24-105mm f/4L IS 

criadoperez:
Canon lens appears as "Canon EF 24-105mm f/4L IS USM" and it's correct.
However it seems than sometimes (maybe different production batches of the same lens) it appears as "Canon EF 24-105mm f/4L IS"
This IS lens ALWAYS has USM, so it is the same lens. I copied the values of "Canon EF 24-105mm f/4L IS USM"  into the new lens "Canon EF 24-105mm f/4L IS" 

Some exiv2 extracts:
Exif.Canon.LensModel                         Ascii      74  EF24-105mm f/4L IS USM
Exif.CanonCs.LensType                        Short       1  Canon EF 24-105mm f/4L IS

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -2005,6 +2005,152 @@
 
     <lens>
         <maker>Canon</maker>
+        <model>Canon EF 24-105mm f/4L IS</model>
+        <mount>Canon EF</mount>
+        <cropfactor>1</cropfactor>
+        <calibration>
+            <distortion model="ptlens" focal="24" a="0.017263" b="-0.049244" c="0"/>
+            <distortion model="ptlens" focal="28" a="0.010878" b="-0.024454" c="0"/>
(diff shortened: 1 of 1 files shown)


MStraeten:
why not using this way instead of a duplicate 
```
        <model>EF35mm f/1.4L II USM (750)</model>
        <model lang="en">Canon EF 35mm f/1.4L II USM</model>
```

criadoperez:
Thanks @MStraeten for that suggestion.
If that is possible, I think it would be preferred. As just wasn't sure it worked that way. I suggested that [in this issue](https://github.com/lensfun/lensfun/issues/1281).
I can resend the PR with that change.

criadoperez:
Closed pull request. New one is [here](https://github.com/lensfun/lensfun/pull/1323).

PR #1290, opened 2020-10-30
Update compact-sony.xml (added RX100-VII entry)

MathiasBayon:
Added RX100 VII, which has exact same lens characteristics as RX100 VI


tuxfanx:
Sony RX100VII was taken over into the DB with PR #1635

PR #1322, opened 2020-12-12
Update mil-fujifilm.xml

dimk1:
Added the latest X-S10. Not sure about the crop factor though, how can I check if it is correct? I entered the same as X-T30 which they are said to have the same APS-C sensor.

### DIFF
--- a/data/db/mil-fujifilm.xml
+++ b/data/db/mil-fujifilm.xml
@@ -131,6 +131,13 @@
 
     <camera>
         <maker>Fujifilm</maker>
+        <model>X-S10</model>
+        <mount>Fujifilm X</mount>
+        <cropfactor>1.534</cropfactor>
+    </camera>
+    
+    <camera>
+        <maker>Fujifilm</maker>
         <model>X-T100</model>
         <mount>Fujifilm X</mount>
         <cropfactor>1.529</cropfactor>
--- a/data/db/mil-fujifilm.xml
+++ b/data/db/mil-fujifilm.xml
@@ -75,6 +75,13 @@
 
     <camera>
         <maker>Fujifilm</maker>
+        <model>X-T4</model>
+        <mount>Fujifilm X</mount>
+        <cropfactor>1.534</cropfactor>
+    </camera>
+    
+    <camera>
+        <maker>Fujifilm</maker>
         <model>X-Pro3</model>
         <mount>Fujifilm X</mount>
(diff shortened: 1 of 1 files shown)


majobu:
can you do the same for X-T4 with the data of X-T3? Don't know how to do that and do not understand what to do....

dimk1:
I did, but I would like some more experienced developer to check my changes.

dbrgn:
The same X-Trans 4 sensor is shared among the X-T3, X-T4, X-T30, X-Pro3, X100V and X-S10. Using the same crop factor as the X-T3 should work.

I'm not sure where the exact value 1.534 comes from.

According to https://www.dpreview.com/forums/thread/3828726:

> Dimensions of a typical full-frame sensor (horizontal, vertical, diagonal, in mm): 35.8, 23.8, 43.0
> Dimensions of the Fuji APS-C sensor (horizontal, vertical, diagonal, in mm): 23.6, 15.6, 28.3
> Crop factors: (horizontal, vertical, diagonal): 1.52, 1.53, 1.52

Not sure what full-frame sensor size lensfun uses as a base for calculations.

@dimk1 maybe you could rename your pull request to "Add support for Fujifilm X-T4 and X-S10" so that it's more clear?

zinga666:
Hi, the X-E4 recently was released and has the same sensor as X-T4 and X-S10 as well. @dimk1 could you probably add that one as well, please? I didn't want to create yet another PR and this fits here perfectly... ;-)

gnl21:
> Not sure about the crop factor though, how can I check if it is correct?

I tend to look up the cameras on https://www.digicamdb.com/. It seems to have pretty good data and it confirms all the cameras in this PR (and the X-E4) have the crop factor 1.534. (The website only displays the crop factor to 1 decimal place, but working it out from the sizes gives the more accurate answer).

> Not sure what full-frame sensor size lensfun uses as a base for calculations.

Lensfun uses 35mm (so called "full frame") as the basis of the crop factor calculations.

@dimk1, it would be great if you could make the commit message slightly more descriptive (something like "Add some Fuji mirrorless cameras" and the models in the body of the commit message would be great) and then I'll merge the PR. If you want to add the X-E4 as zinga666 requested that would be great, otherwise I can do it separately.

dbrgn:
@gnl21 could you squash-merge the two commits and update the commit message there, to avoid blocking the merging of this change? 🙂 The additional camera models could be added in separate PRs.

Issue #1281, opened 2020-10-22
Canon lens with double naming


criadoperez:
Canon lens appears as "Canon EF 24-105mm f/4L IS USM" and it's correct.
However it seems than sometimes (maybe different production batches of the same lens) it appears as "Canon EF 24-105mm f/4L IS"
This IS lens ALWAYS has USM, so it is the same lens. I have copied the values of "Canon EF 24-105mm f/4L IS USM"  into the new lens "Canon EF 24-105mm f/4L IS" 

Some exiv2    ###extracts:
Exif.Canon.LensModel                         Ascii      74  EF24-105mm f/4L IS  #USM
Exif.CanonCs.LensType                        Short       1  Canon EF 24-105mm f/4L IS # #1280 

Pull request #1280

criadoperez:
Instead of adding a new lens, maybe we could also solve the issue but just adding the following line to the current lens:
`<model lang="en">Canon EF 24-105mm f/4L IS</model>`
Is this how it works?
If this is the preferred method, please let me know and I will make that change instead with a new pull request.

tuxfanx:
lens recognition should work for both. "EF 24-105 mm 4.0 L IS USM" is the correct name for the lens, so "<model lang="en">Canon EF 24-105mm f/4L IS</model>" makes no sense --> closed

PR #1323, opened 2020-12-14
Update slr-canon.xml. Added secondary name to lens

criadoperez:
[Corrects Issue 1281](https://github.com/lensfun/lensfun/issues/1281). 

Canon lens appears as "Canon EF 24-105mm f/4L IS USM" and it's correct.
However it seems than sometimes (maybe different production batches of the same lens) it appears as "Canon EF 24-105mm f/4L IS"
This IS lens ALWAYS has USM, so it is the same lens. I copied the values of "Canon EF 24-105mm f/4L IS USM" into the new lens "Canon EF 24-105mm f/4L IS"

Some exiv2 extracts:
Exif.Canon.LensModel Ascii 74 EF24-105mm f/4L IS USM
Exif.CanonCs.LensType Short 1 Canon EF 24-105mm f/4L IS

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -1860,6 +1860,7 @@
     <lens>
         <maker>Canon</maker>
         <model>Canon EF 24-105mm f/4L IS USM</model>
+        <model lang="en">Canon EF 24-105mm f/4L IS</model>
         <mount>Canon EF</mount>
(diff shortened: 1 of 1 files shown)


gnl21:
As I understand the lensfun searching, having extra elements like "USM" at the end of the name shouldn't prevent it from matching. Does this change solve a problem that you were having?

PR #1330, opened 2020-12-22
Adding Canon EOS 850D/Rebel T8i/Kiss X10i

SebastianPfliegel:
Adding Canon's latest DSLR


gnl21:
Thanks for this. Looks good to me.

PR #1343, opened 2021-01-08
Fix model for Sigma 30mm f/1.4 DC DN C 016

allgoewer:
Hi,

darktable did not detect the lens automatically, it's reported in darktable with "Contemporary" spelled out. I simply changed the model in mil-sigma.xml .

**Before the patch** 
darktable_lens_not_detected

**After the patch** 
dartable_lens_fixed

### DIFF
--- a/data/db/mil-sigma.xml
+++ b/data/db/mil-sigma.xml
@@ -184,7 +184,7 @@
 
     <lens>
         <maker>Sigma</maker>
-        <model>SIGMA 30mm f/1.4 DC DN | C 016</model>
+        <model>SIGMA 30mm f/1.4 DC DN | Contemporary 016</model>
         <model lang="en">30mm f/1.4 DC DN</model>
         <mount>Sony E</mount>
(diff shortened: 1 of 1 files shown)


tgurr:
I remember having a similiar issue with the calibration data for the 16mm with darktable and RawTherapee, see my comment back then: https://github.com/lensfun/lensfun/issues/544#issuecomment-510815875 and @bronger gave me the solution in a comment of the PR: https://github.com/lensfun/lensfun/pull/939#issuecomment-512519246

So maybe the following would be a more compatible approach: 

```
<maker>Sigma</maker>
<model>Sigma 30mm f/1.4 DC DN | Contemporary C 016</model>
<model lang="en">Sigma 30mm f/1.4 DC DN C</model>
```

allgoewer:
@tgurr Thanks for the suggestion!

cjelgin:
Thanks for doing this guys. I've been having this same issue in darktable and came here to get it fixed.

PR #1351, opened 2021-01-26
Match library file name with soname

mikhailnov:
Before this change there was a symlink liblensfun.so.2 -> liblensfun.so.0.3.95
The soname was liblensfun.so.2
File name must match the soname


seebk:
Are you sure your change is correct? The SOVERSION is the binary 
interface version while the VERSION is the actual library version.

https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html

Lensfun has not always incremented the VERSION_MAJOR together with 
VERSION_API in the past. That's why the latter already is at 2 but 
VERSION_MAJOR is still 0.

mikhailnov:
>  Are you sure your change is correct? The SOVERSION is the binary
> interface version while the VERSION is the actual library version.
>
> https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html
>
> Lensfun has not always incremented the VERSION_MAJOR together with
> VERSION_API in the past. That's why the latter already is at 2 but
> VERSION_MAJOR is still 0.
In most projects ABI is versioned separately from the project itself, just as in this case, but I've _never_ seen a library which's file name does not match the soname.

seebk:
I still think our naming scheme is correct but will leave this issue open and move it to the Stable Release milestone to get further opinions from other devs

PR #1365, opened 2021-02-24
Update slr-pentax.xml

gribouilleuse:
"Pentax K-1" and "Pentax K-1 mark II" have a KAF4 mount and is compatible with all old version of K mount. The main difference between KAF3 and KAF4 is electromagnetic diaphragm control system on KAF4.


sboukortt:
Merci !

PR #1376, opened 2021-03-04
Add Canon EF-M 32mm f/1.4 lens

huanghai2333:
calibration for distortion has been tested


gnl21:
This seems to have an order of magnitude more distortion than a similar EF-M prime that's already in the database and reviews seem to say this lens has minimal distortion (although I can't find one that measured it). Is this correction giving good results for you?

PR #1394, opened 2021-03-27
Add Sony FE 24mm F1.4 GM

MikkoTervala:
Add the Sony FE 24mm F1.4 GM E mount lens.
Distortion, TCA and vignetting.
Images taken with Sony a7C ILCE-7C

TCA measurements could be redone with a higher resolution body.
Otherwise I'm very satisfied with the results.

[lensfun_calibration Sony 24mm.zip](https://github.com/lensfun/lensfun/files/6216011/lensfun_calibration.Sony.24mm.zip)


sarunasb (commented):
Shouldn't maker element be at the same indent as model?

tuxfanx:
> Shouldn't maker element be at the same indent as model?

Definitely, it was the first time I edited a merge conflict - but things it went wrong. I'm not that familiar with the process. But now everything should be correct.

PR #1406, opened 2021-04-21
Add Fuji X-E4

gnl21:


gnl21:
@zinga666 I think this is the additional Fuji camera that you asked for. I'll merge this shortly, but if you have a chance to test it out that would be great.

zinga666:
@gnl21 Yes! :-) Unfortunately no - I don't have the possibility to try it out. Just planning to buy that camera in the future and currently trying to get support into darktable, etc.

gnl21:
Fair enough. I don't think there's a problem with merging this now, so I'll do that. Happy saving!

PR #1421, opened 2021-05-22
Bump version to 2.0.0

payano:
This is a way for darktable to use the latest version of the liblensfun library.
See: https://github.com/darktable-org/darktable/pull/9037


elstoc:
darktable is unlikely to make any changes to support the latest lensfun until #945 is resolved. Even then it will be somewhat problematic since some distributions are already using 0.3.95 (which also exhibits this fault).

payano:
Okay, 
It's not that good that some distros use the 0.3.95 version though, they might have issues stepping up to a newer version.

I read that @uchrisu had a fix for that issue, is that still a viable solution? I'm just interested on how we can proceed and solve this issue. As far as I've read there has been some improvements in the code base, and it would be nice to step up the version and let everyone use it.

seebk:
Sorry, but I think we should not move to version 2.0.0 until all or at least most important ones from the open issues in the Stable Release Milestone are fixed. 

 Furthermore this PR contains too many changes apart from bumping the version number in the commit list so I am closing it now.

PR #1422, opened 2021-05-26
Adding EOS R, RP, R5, R6

alexvanderberkel:
The new EOS R series has a couple of cameras which is not listed in the list of cameras. This PR adds 4 new cameras to the list including the a new mount type RF.

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -188,6 +188,39 @@
 
     <camera>
         <maker>Canon</maker>
+        <model>Canon EOS R</model>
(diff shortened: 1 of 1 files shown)


gnl21:
This commit has broken the autobuilds because these cameras actually are in the database, so I'm going to merge a PR reverting this change. If you update to the latest database, are you still having a problem with lensfun for these cameras? (They're listed in mil-canon.xml).

alexvanderberkel:
I will take a look later this evening but please go ahead with reverting my change. Sorry for the inconvenience.

PR #1442, opened 2021-06-21
Use official Canon RF lens model names

cytrinox:
These are the official lens names for Canon RF.
exiftool already uses these names and a patch for exiv2 is prepared.

### DIFF
--- a/data/db/mil-canon.xml
+++ b/data/db/mil-canon.xml
@@ -584,7 +584,7 @@
 
     <lens>
         <maker>Canon</maker>
-        <model>RF24-105mm F4 L IS USM</model>
+        <model>Canon RF 24-105mm F4L IS USM</model>
         <mount>Canon RF</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>
@@ -692,7 +692,7 @@
 
     <lens>
         <maker>Canon</maker>
-        <model>RF70-200mm F2.8 L IS USM</model>
+        <model>Canon RF 70-200mm F2.8L IS USM</model>
         <mount>Canon RF</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>
@@ -709,7 +709,7 @@
     
     <lens>
         <maker>Canon</maker>
-        <model>Canon RF85mm F1.2L USM</model>
+        <model>Canon RF 85mm F1.2L USM</model>
         <mount>Canon RF</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>
@@ -740,7 +740,7 @@
 
     <lens>
         <maker>Canon</maker>
-        <model>RF85mm F2 MACRO IS STM</model>
+        <model>Canon RF 85mm F2 MACRO IS STM</model>
         <mount>Canon RF</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>


cytrinox:
@matt-maguire Patch for exiv2 is here: https://github.com/Exiv2/exiv2/pull/1729

gnl21:
Since this has been merged into exiv2 I guess I'll merge this, but I'm a bit worried that it will cause compatibility problems for people who have edited images using a version of exiv2 prior to this fix. Do you know whether an exiv2 release has gone out with the incompatible names in it?

The autobuilds were broken on master recently, so could you update the PR to the latest code to pick up the fix?

cytrinox:
Afaik exiv2 0.27.4 was released with wrong names. But as RF lenses can only be attached to Canon cameras writing CR3 raws and CR3 support is very limited for now, I assume there are not many people out there dealing with lensun + RF lenses.

matt-maguire:
This will cause breakage -- not everyone will move to the new exiv2 immediately. You should define the existing names as an alias to avoid this.

gnl21:
I think that what we should do is change the names as proposed in this PR, but include
```
<model lang="en">Old (incorrect) name</model>
```
as well. I think that will give us the "official" Canon names as the names in the database but will also allow the old names to keep matching.

Slightly OT, but do you know of anywhere with Ubuntu packages of exiv2 0.27.4 with ISOBMFF support? I can't find any and that would allow me to test this change relatively easily, rather than waiting until I have time to build it from source.

cytrinox:
```
 <model lang="en">Old (incorrect) name</model>
 ```
Is this already supported? Then I will update this PR in the next weeks (currently working on another project & holiday is near).

cytrinox:
Requested changes are done since 3 months. Are there more changes required to get this merged?

sarunasb:
> ```
> <model lang="en">Old (incorrect) name</model>
> ```
> Is this already supported? Then I will update this PR in the next weeks (currently working on another project & holiday is near).
Matching is done using `<model>` lines, not `<model lang="">` ones.

How the changes have been tested?

gnl21:
When I suggested the change I was intending to test it, at least using the 1 RF lens that I have. Since then I haven't really had time to do any work on lensfun, but I'll try to verify this in the next few days and merge it.

gnl21:
> Matching is done using <model> lines, not <model lang=""> ones.

Ah, OK, I thought it was done using all the defined names. That was one of the things that I was going to test, but if you know that it doesn't work then I don't think it's really possible to define these names as aliases at the moment.

So, I've tested this change using exiv2 release 0.27.4 and it already appears to include this change, even though I can't find the linked PR in the git log. So, anyone using any release of exiv2 (rather than having compiled it themselves before 0.27.4 was finished) should have these correct names. I can say that the OBS Ubuntu builds for darktable never packaged a version for which this PR would not work.

Looking through the issues linked from this one, it's also not clear to me that this lens matching was ever working for these lenses (although the posters there don't say which version of exiv2 they were using).

So if we want to preserve 100% backwards compatibility, based on what @sarunasb says, I think we'd have to have 2 copies in the DB with the same data but different names. Given that these packages have not hit distros yet, there's an easy upgrade to exiv2 0.27.5 that will fix things for anyone affected, and there may never have been an actual release with the bad data in, should we just merge this PR? Does anyone have an alternative suggestion? Is there anyone who has had this working with the existing database and a previous release of exiv2 who can comment on which users, if any, would be broken by this change?

piratenpanda:
Indeed it seems like the changes from https://github.com/Exiv2/exiv2/pull/1729/files are in the maintainance branch, without a commit that changed it. See https://github.com/Exiv2/exiv2/blob/8968ceb166713c8b686faafb402df4eb41720ed5/src/canonmn_int.cpp#L2052

I think it's also save to assume that anyone using cr3 files or those lenses uses at least exiv2 v0.27.4. Would anyone really be affected by this PR who does run older eviv2 versions. Do some distros backport BMFF support? That would be kinda crazy to me.

gnl21:
OK, I think I've solved what's going on here. I had thought that this change was making the same updates as the exiv2 change but it's not, the existing names in lensfun are broken with both the old and the new exiv2 names.

The "Canon" that is added in this change is needed whether you have the exiv2 update or not, so it doesn't matter whether the remaining part matches a particular exiv2 release. I'm pretty sure that the existing names can't have been matching properly for anyone, so there's nothing to break by changing them. (And also no need for the aliases -- sorry about that.)

I don't have any of these lenses but I've tested by making analogous changes for the RF lens that I do have and the new names appear to work for me (although, weirdly, the spaces in the name darktable returns don't match this change or the exiv2 source code).

TL;DR I think we should accept this change. If @cytronix wants to rebase it to resolve conflicts and, optionally, remove the aliases then, absent other objections, I'll merge this PR. If not (and let's face it, we've messed him around long enough already), then I can do it. If there are still any objections then please say now (and I'd particularly like to hear from @matt-maguire, in light of his comment above).

cytrinox:
@gnl21 Rebased, removed alternative name and fixed some more model names added recently.

gnl21:
That's excellent, thanks. I noticed the new ones and was going to fix it later, but this is even better.

I'll leave this open a little longer so that anyone who wants to object has a chance to (it's the middle of the night in Australia right now), but if there are no objections then I'll merge this tomorrow at the latest.

cytrinox:
Thanks. I've the RF100 f/2.8 and RF 15-35 f/2.8 in pipeline, so if this gets merged I open PRs for these lenses.
@piratenpanda The RF 14-25 f/4 is a bit problematic, the vignetting correction is horrible because of the construction of the lens. I need to find a solution for this until I make a PR for this lens.

piratenpanda:
Most likely it doesn't matter anyway as this gets removed by distortion correction?

gnl21:
The autobuilds look quite broken but none of it looks related to this change. The debug configurations all time out (which seems common to all PRs), most of the release configurations work except that one configuration failed because glib was already installed. I'll merge this now and look at fixing the autobuilds separately.

PR #1450, opened 2021-07-06
Remove Travis CI from readme.md

payano:
Travis CI is not in use anymore.


payano:
Isn't it good to have Travis CI which builds the library for Linux? I thought that Appveyor builds the window version?

gnl21:
Looking at the logs from the build on this change, I would say that yes, AppVeyor is only building on windows.

I would also say that yes, it's a good idea to have some build running on linux. I don't know anything about CI tools that integrate with GitHub and I don't know why the old Travis CI thing got turned off. If we can start it running again then that would be great, but otherwise I guess I'll merge this PR.

payano:
I think we can remove Travis ci and then use the github build system to build the linux library, I'll try to fix the configuration ASAP.

seebk:
Thanks! It seems Travis now has limited ressources for free users and we exceeded our credits. Agree we shoul dmove to Github Actions.

Issue #1353, opened 2021-02-03
Add support for Nikon Z 7 II


PeterWem:
```
<camera>
        <maker>NIKON CORPORATION</maker>
        <maker lang="en">Nikon</maker>
        <model>NIKON Z 7_2</model>
        <model lang="en">Nikon Z7 II</model>
        <mount>Nikon Z</mount>
        <cropfactor>1</cropfactor>
    </camera>
```

Sturmflut:
Fixed in lensfun@e6cd7aabdf2de44b0fb51d0a7bf8349a99d437be

Issue #1371, opened 2021-03-03
Add Support for Nikon Z6_2


talmholt:
<camera>
        <maker>Nikon Corporation</maker>
        <maker lang="en">Nikon</maker>
        <model>Nikon Z6_2</model>
        <model lang="en">Z6_2</model>
        <mount>Nikon Z</mount>
        <cropfactor>1</cropfactor>
    </camera>

Sturmflut:
Fixed in lensfun@e6cd7aabdf2de44b0fb51d0a7bf8349a99d437be

PR #1483, opened 2021-09-23
Clean up some Nikon MILC strings for readability

kmilos:
Should have no effect on matching (case and spaces).

Z 6 and Z 7 models actually have a space in the Exif data as set by the manufacturer.

Also added new camera models and fixed some indents.

### DIFF
--- a/data/db/mil-nikon.xml
+++ b/data/db/mil-nikon.xml
@@ -129,8 +129,8 @@
     <camera>
         <maker>Nikon Corporation</maker>
         <maker lang="en">Nikon</maker>
-        <model>Nikon Z6</model>
-        <model lang="en">Z6</model>
+        <model>Nikon Z 6</model>
+        <model lang="en">Z 6</model>
         <mount>Nikon Z</mount>
         <cropfactor>1</cropfactor>
     </camera>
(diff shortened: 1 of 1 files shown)


freepenguin84 (approved):
Made these changes locally since I have a Nikon Z 6_2 and wondered why it wouldn't be recognized. Works!

Sturmflut:
I agree with the camera model names. It is confusing that "Z6" and "Z7" did not initially have the spacing and Nikon marketing decided to retroactively add it in all the official materials, but hey, they sell the products.

I would merge this, since it also fixes #1353 and #1371. But my issue with this pull request is that it also renames the "NIKKOR" in some lenses to "Nikkor", though. Nikon is not fully consistent with the uppercase/lowercase naming, but the vast majority of FX and DX lenses has the fully uppercase NIKKOR, with "Nikkor" seemingly only being used in combined names that have a dash (like "Zoom-Nikkor" or "Micro-Nikkor", but not "Micro NIKKOR") and for some older manual focus lenses: https://imaging.nikon.com/lineup/lens/f-mount/

The Z lineup uses the fully-uppercase "NIKKOR" without exception: https://imaging.nikon.com/lineup/lens/z-mount/

I was about to add the Nikon D6, D780 and Z 9, and there are two other open issues for the Z 6II and Z 7II. I'll collect everything in [/Sturmflut/lensfun/tree/2021-01-nikon-fixes](/Sturmflut/lensfun/tree/2021-01-nikon-fixes) and merge that one.

Sturmflut:
Merged, closing all issues and this pull request

kmilos:
Thanks

kmilos:
Btw, I really think Z 6 and Z 7 should have a space in the model name, as identified in the Exif metadata, rather than any marketing material...

Sturmflut:
Yes, I had a look at some sample RAWs. Fixed in lensfun@9b8009e0a8631ddb01c411cd075c601939101b48

PR #1508, opened 2021-11-18
Fixed incorrect return

dkollmann:
I fixed an incorrect return in lfModifier::EnableProjectionTransform and fixed some compiler warnings with VS2017.


seebk:
Fixed by 29c6f4609f28.

PR #1316, opened 2020-12-05
Adding X100V, XF16-80mm f4

t3mujin:
Cameras
- X100V
- X100F

Lens
- XF16-80mmF4 R OIS WR (tca, vignetting and distortion)
- X100V (tca, vignetting and distortion), and since the x100v has a new lens I created a new lens mount (x100v) for this and upcoming cameras of this family.

PR #1525, opened 2021-12-09
Cherrypick settings for Fujifilm XF16-80mmF4 R OIS W from #1316

Carsten-Bruch:
Taking over settings for  Fujifilm XF16-80mmF4 R OIS W from #1316 to solve CI failures


tuxfanx:
xf 16-80mm is already in DB, I merged the vignetting data from this PR with PR #1632

PR #1543, opened 2021-12-23
Fix name of Tamron 17-35mm

Leonidas-from-XIV:
When selecting this in e.g. darktable it includes the maker name. Also, it is consistent with the other Tamron names in this file.


Leonidas-from-XIV:
Thanks!

PR #1550, opened 2021-12-27
Add calibration data for the RF 100-500mm F4.5-7.1 L IS USM 

piratenpanda:
and a minor distortion update for the RF 85 1.2


piratenpanda:
Some values seem to be wrong, will do another one

Issue #1554, opened 2022-01-01
Calibration upload 42598a


bronger:
Calibration images were uploaded to the directory that starts with “`42598a_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

guyru:
Hi, I'm the uploader. These are test shots for Tamron 18-300mm F3.5-6.3 B061 for Sony

guyru:
I've performed the calibration and opened a pull request: https://github.com/lensfun/lensfun/pull/1555

PR #1555, opened 2022-01-07
Add Tamron 18-300mm F3.5-6.3 for Sony E

guyru:
Distortion correction for Tamron 18-300mm F3.5-6.3 for Sony E


tuxfanx:
checked with pictures from upload #1554 in darktable

PR #1556, opened 2022-01-07
Add vignetting data for Nikon AF-S Nikkor 200-500mm f/5.6E ED VR

Sturmflut:


sarunasb:
@Sturmflut Vignetting correction already exists for 200-500mm f/5.6E ED VR. Is this a tested improvement?

PR #1557, opened 2022-01-07
Add vignetting and TCA data for Nikon Nikkor AF-S 24-70mm F2.8G ED

Sturmflut:


tuxfanx:
Correction data are completely contained in the database.

PR #1558, opened 2022-01-07
Add vignetting data for Nikon AF-S Nikkor 16-35mm f/4G ED VR

Sturmflut:


tuxfanx:
vignetting data are already in DB. Sadly I don't have any photos to compare

PR #1574, opened 2022-01-11
Add distortion and tcs data for Sigma 18-50mm f2.8 DC DN

swidnow2:
Calibration directory: 830c62_kedzierski.damian


tuxfanx:
The APS-C lens is also available with an L-mount. The model name for the L-mount is the same. Maybe you could add it and then fix the merge conflict right away?

tuxfanx:
see PR #1646

Issue #1511, opened 2021-11-19
Camera Panasonic DC-S5


pahervin:
Please add the Panasonic DC-S5 camera :

- crop factor = 1
- mount = Leica L

Sturmflut:
The linked pull request has the data.

Sturmflut:
Merged into master.

PR #1576, opened 2022-01-12
Add Panasonic DC-S5 camera (#1511)

Sturmflut:


Sturmflut:
Fixes #1511

Sturmflut:
EXIV data verified against public RAW sample image https://www.imaging-resource.com/PRODS/panasonic-s5/S5hSLI000050NR1.RW2.HTM

PR #1588, opened 2022-01-18
add another name for the sigma 150-600mm contemporary

### LINKED ISSUE #1586: "Sigma 150-600 F5-6.3 DG OS HSM Contemporary" not identified properly
I have a `Sigma 150-600 F5-6.3 DG OS HSM Contemporary` that I use with an FTZ adapter on a Nikon Z6. This combination identifies the lens as `Sigma 150-600 F5-6.3 DG OS HSM Contemporary` in Darktable, but in the Lensfun database it is identified as `Sigma 150-600 f/5-6.3 DG OS HSM | C`, and thus doesn't match automatically. On a native F-mount body (Nikon D5300) it also uses this name, so this may be due to a change in exiv2 / darktable instead of the FTZ adapter as I initially suspected.

I don't know if this is true on other mounts as well, but I think this name should at least be added as an alternate.

benklop:
This resolves #1586 for me. I don't know if this is necessary for mounts other than F mount, but i think it's likely.

### DIFF
--- a/data/db/slr-sigma.xml
+++ b/data/db/slr-sigma.xml
@@ -3421,6 +3421,7 @@
     <lens>
         <maker>Sigma</maker>
         <model>Sigma 150-600mm f/5-6.3 DG OS HSM | C</model>
+        <model lang="en">Sigma 150-600mm F5-6.3 DG OS HSM Contemporary</model>
         <mount>Canon EF</mount>
         <mount>Sigma SA</mount>
         <mount>Nikon F AF</mount>
@@ -3530,6 +3531,7 @@
     <lens>
         <maker>Sigma</maker>
         <model>Sigma 150-600mm f/5-6.3 DG OS HSM | C</model>
+        <model lang="en">Sigma 150-600mm F5-6.3 DG OS HSM Contemporary</model>
         <mount>Canon EF</mount>
         <mount>Sigma SA</mount>
         <mount>Nikon F AF</mount>
@@ -3558,6 +3560,7 @@
     <lens>
         <maker>Sigma</maker>
         <model>Sigma 150-600mm f/5-6.3 DG OS HSM | C</model>
+        <model lang="en">Sigma 150-600mm F5-6.3 DG OS HSM Contemporary</model>
         <mount>Canon EF</mount>
         <mount>Sigma SA</mount>
         <mount>Nikon F AF</mount>


sarunasb:
What matters is what 
`exiv2 -pt yourRAWfile|grep lens`
reports. That string should go into `<model>` element. `<model lang=''>` is for your eyes only.

benklop:
> exiv2 -pt yourRAWfile|grep lens

This returns nothing, but on the D5300 grepping for `Lens` returns 

```
Exif.Nikon3.LensType                         Byte        1  D G VR
Exif.Nikon3.Lens                             Rational    4  150-600mm F5-6.3
Exif.Nikon3.LensFStops                       Undefined   4  4.33333
Exif.NikonLd3.LensIDNumber                   Byte        1  Sigma 150-600mm F5-6.3 DG OS HSM Contemporary
Exif.NikonLd3.LensFStops                     Byte        1  F4.3
```
I assume this means `Sigma 150-600mm F5-6.3 DG OS HSM Contemporary` is the correct model value then, and the existing value should simply be changed instead of added to?

kmilos:
This should indeed be correct, as darktable [tries to figure out if an FTZ adapter is used](https://github.com/darktable-org/darktable/blob/6ef3f6b0583b77eedaf240bed58b28d8a6b900a0/src/common/exif.cc#L1064-L1075) and then prefer the exiv2 name lookup like it is done for all other F mount lenses.

Just out of curiosity, what is `Exif.Photo.LensModel` w/ FTZ, how come it is not printed above?

kmilos:
Indeed, this is "Sigma 150-600mm f/5-6.3 DG OS HSM | C" in exiv2 for Canon 😖 

So best not to touch it after all but handle this w/ your local ~/.exiv2 (or exiv2.ini on Windows) mechanism for lens remapping?

benklop:
> Indeed, this is "Sigma 150-600mm f/5-6.3 DG OS HSM | C" in exiv2 for Canon confounded
> 
> So best not to touch it after all but handle this w/ your [local ~/.exiv2 (or exiv2.ini on Windows) mechanism for lens remapping](https://exiv2.org/manpage.html)?

I've certainly done this, but that is m i thought, intended to be a temporary workaround to allow getting on with my life while the _real_ fix gets applied upstream. do you really want to discourage people from contributing metadata fixes??

benklop:
actually, maybe the correct answer is to update exiv2's canon definition to make it consistent, then fix it here as well?

benklop:
> Just out of curiosity, what is `Exif.Photo.LensModel` w/ FTZ, how come it is not printed above?

I have no idea - all I can tell you is that is what's there.

kmilos:
> do you really want to discourage people from contributing metadata fixes?

> maybe the correct answer is to update exiv2's canon definition to make it consistent

Or Nikon's? Which one is the reference? Lensfun? Hardly...

Therein lies the rub, and in a way, yes, we want to discourage this, because the "fix" is a very vague and relative term here. This is because there is no unambiguous way to do this. The MakerNote LensID to lens name lookup is completely arbitrary, and as you have found out differs between exiftool, exiv2, and mounts; the naming scheme even differs within the same same app for the same mount. The strings we have today are contributed by users and are in no way authoritative. Not even lens manufactures use consistent names in their own marketing material and product specs.

Only w/ RF and Z mount did manufacturers start putting the authoritative lens string in the standard Exif LensModel field made for this purpose and we can talk about fixes. (This is for Canon and Nikon; some other systems might be using the Exif LensModel for longer already.)

uchrisu:
For reference: 
The Canon lens adapted to Sony-E gives:
Exif.Photo.LensModel : Ascii : 46 : 150-600mm F5-6.3 DG OS HSM | Contemporary 015

kmilos:
Thanks @uchrisu But again, I presume this is Sony's body firmware interpretation, not an actual string defined by Sigma's lens firmware?

kmilos:
Btw, does adding the different model alias rather than replacing perhaps work, like it was attempted [here](https://github.com/lensfun/lensfun/pull/1039)?

sarunasb:
Adding a `<model lang` line will not affect autodetection.

kaesa:
exiv2 -pt *rawfile* | grep Lens, done for the EF mount version

This is the results from Canon 2000D CR2 file:
```
Exif.CanonCs.LensType                        Short       1  Unknown Lens (624)
Exif.CanonCs.Lens                            Short       3  150.0 - 600.0 mm
Exif.Canon.LensModel                         Ascii      70  150-600mm F5-6.3 DG OS HSM | Contemporary 015
Exif.CanonLe.LensSerialNumber                Byte        5  0000000000
Exif.Photo.LensSpecification                 Rational    4  150-600mm
Exif.Photo.LensModel                         Ascii      46  150-600mm F5-6.3 DG OS HSM | Contemporary 015
Exif.Photo.LensSerialNumber                  Ascii      12  0000000000
```

This is the result from Canon EOS R10 CR3 file with Canon EF->RF adapter:
```
Exif.Photo.LensSpecification                 Rational    4  150-600mm
Exif.Photo.LensModel                         Ascii      46  150-600mm F5-6.3 DG OS HSM | Contemporary 015
Exif.Photo.LensSerialNumber                  Ascii      11  0000000000
Exif.CanonCs.LensType                        Short       1  Unknown Lens (624)
Exif.CanonCs.Lens                            Short       3  150.0 - 600.0 mm
Exif.CanonFi.RFLensType                      SShort      1  n/a
Exif.Canon.LensModel                         Ascii     138  150-600mm F5-6.3 DG OS HSM | Contemporary 015
Exif.CanonLiOp.DigitalLensOptimizer          SLong       1  Standard
Exif.CanonLe.LensSerialNumber                Byte        5  0000000000
Exif.CanonAfC.USMLensElectronicMF            SLong       1  Disable After AF
Exif.CanonAfC.LensDriveWhenAFImpossible      SLong       1  Continue Focus Search
```

sarunasb:
So to ensure matching for all systems the `model` line then should be:

`<model>Sigma 150-600mm f/5-6.3 DG OS HSM | Contemporary 015</model>`

If that's too much for the eyes, the human readable version can be prettyfied by adding:

`<model lang="en">Sigma 150-600mm f/5-6.3 DG OS HSM | C</model>`

PR #1590, opened 2022-01-18
 Add Lumix DC-FZ1000 II camera, same lens as FZ-1000, update compact-panasonic.xml

greenoid:
Add Panasonic Lumix DC-FZ1000 II. Same lens as its predecessor DC-FZ1000.


Sturmflut:
Looks correct, sensor and lens don't seem to have changed. Merging.

PR #1591, opened 2022-01-19
Add Fujifilm XC 35mm F2 Support

### LINKED ISSUE #1180: How can I create a new lens profile from an existing profile to access it from Darktable?
I have no knowledge how to create a new lens profile for lensfun that can be accessed from Darktable under Windows. I searched in Internet several sites, but this subject seems to be complicated. I could not find any straightforward guide explaining the procedure step by step.

What I need is the following:
I own a newer Fujifilm lens XC 35 mm f/2.0 for that there exists no entry in Darktable for lens correction. I noticed there is another Fujifilm lens XF 35 mm f/2.0 that I can select manually and the result of correction is reasonable. Both lenses may have a very similar construction.

I would like to create a lens profile (I assume it is a XML text file) for XC 35 mm f/2.0 just by importing the existing data of the lens XF 35 mm f/2.0.  

My first question is in which folder Lensfun database under Windows 8.1 is stored, when I install Darktable. How are the lens profiles named and how I can create a new one so that it can be automatically recognized by Darktable.

Thank you for your help in advance.
Regards,
Mustafa K.

noahwc:
Fixes #1180. As noted in the issue the XC 35mm F2 and XF 35mm F2 WR appear to be optically identical. This is confirmed in my own testing and the lens data sheets [1][2]. Therefore the new profile is identical to the existing XF 35mm F2 WR profile.

[1] https://fujifilm-x.com/global/products/lenses/xf35mmf2-r-wr/specifications/
[2] https://fujifilm-x.com/global/products/lenses/xc35mmf2/specifications/

### DIFF
--- a/data/db/mil-fujifilm.xml
+++ b/data/db/mil-fujifilm.xml
@@ -1460,7 +1460,7 @@
     <lens>
         <maker>Fujifilm</maker>
         <model>XF35mmF2 R WR</model>
-        <model lang="en">XF 35mm f/2 WR</model>
+        <model lang="en">XF 35mm f/2 R WR</model>
         <mount>Fujifilm X</mount>
         <cropfactor>1.528</cropfactor>
         <calibration>
--- a/data/db/mil-fujifilm.xml
+++ b/data/db/mil-fujifilm.xml
@@ -1488,6 +1488,35 @@
 
     <lens>
         <maker>Fujifilm</maker>
+        <model>XC35mmF2</model>
+        <model lang="en">XC 35mm f/2</model>
+        <mount>Fujifilm X</mount>
+        <cropfactor>1.528</cropfactor>
+        <calibration>
+            <!-- Taken with Fujifilm X-Pro2 -->
+            <distortion model="ptlens" focal="35" a="0.00956" b="-0.07057" c="0.07334"/>
+            <!-- Taken with Fujifilm X-E2 -->
+            <tca model="poly3" focal="35" vr="1.0000774" vb="0.9999724"/>
+            <!-- Taken with Fujifilm X-T20 -->
+            <vignetting model="pa" focal="35.0" aperture="2.0" distance="10" k1="-1.3696769" k2="1.4383322" k3="-0.7466834"/>
+            <vignetting model="pa" focal="35.0" aperture="2.0" distance="1000" k1="-1.3696769" k2="1.4383322" k3="-0.7466834"/>
+            <vignetting model="pa" focal="35.0" aperture="2.8" distance="10" k1="-0.2166397" k2="-0.0858889" k3="-0.1293128"/>
+            <vignetting model="pa" focal="35.0" aperture="2.8" distance="1000" k1="-0.2166397" k2="-0.0858889" k3="-0.1293128"/>
+            <vignetting model="pa" focal="35.0" aperture="4.0" distance="10" k1="-0.3323595" k2="0.7726981" k3="-0.7955168"/>
+            <vignetting model="pa" focal="35.0" aperture="4.0" distance="1000" k1="-0.3323595" k2="0.7726981" k3="-0.7955168"/>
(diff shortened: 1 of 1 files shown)


Sturmflut:
I think it might be enough to add a second <model/> tag to the existing entry, like

```
<model>XF35mmF2</model>
<model>XF35mmF2 R WR</model>
<model lang="en">XF 35mm f/2</model>
```

Could you please try that?

noahwc:
So it looks like
```
<model>XC35mmF2</model>
<model>XF35mmF2 R WR</model>
<model lang="en">XF 35mm f/2 WR</model>
```
doesn't detect the XC35mmF2 as a supported lens while
```
<model>XF35mmF2 R WR</model>
<model>XC35mmF2</model>
<model lang="en">XF 35mm f/2 WR</model>
```
does, although I assume it'd fail for the XF35mmF2 R WR lens. Also in the working case, the matched lens is called "XF 35mm f/2 WR" although I assume this is expected.

The tests were conducted in darktable. Happy to test elsewhere, darktable was just what I had handy.

sarunasb:
IMO, since those are two different lenses (even if optically same), each should have its own `<lens>` element.

Then, assuming ‘XC35mmF2” and ‘XF35mmF2 R WR’ are what's reported by exiv2, I suggest `<model>` line for each might be:

`<model>XC 35mm f/2</model>`

`<model>XF 35mm f/2` R WR</model>

And then foregoing the <model lang=...> lines, since they don't add anything.

noahwc:
Confirmed exiv2 reports `XC35mmF2` and `XF35mmF2 R WR` for each lens. Updated `XF 35mm f/2 WR` -> `XF 35mm f/2 R WR` which also makes it more consistent with other Fujifilm lenses in the database. 

> And then foregoing the lines, since they don't add anything.

By this do you mean each `<lens>` element should only have one `<model>` line? Having a single model line

```
   <model>XF 35mm f/2 WR</model>
```
doesn't seem to work for me when tested with darktable (camera/lens not found error). All other Fujifilm lenses also seem to use two model lines.

sarunasb:
@noahwc Do you have a RAW file for XC 35mm, which you could share?

noahwc:
Definitely, here's a sample:
https://drive.google.com/file/d/1p7mne13_mlKalFk6PCnnIcXIB_ZktL4_/view?usp=sharing

This file is licensed [Attribution 4.0 International (CC BY 4.0)](https://creativecommons.org/licenses/by/4.0/).

sarunasb:
@noahwc Thank for the sample image.

I have two RAW images, one for each of the 35mm f/2 lenses. exiv2 reports:

    $ exiv2 -pa * |grep -i model
    DSCF0275.RAF          Exif.Image.Model                             Ascii       5  X-T1
    DSCF0275.RAF          Exif.Photo.LensModel                         Ascii      64  XC35mmF2
    DSCF3478.raf          Exif.Image.Model                             Ascii       5  X-T1
    DSCF3478.raf          Exif.Photo.LensModel                         Ascii      64  XF35mmF2 R WR

I cloned `<lens>` element for XF35mmF2 R WR and changed `<model>` lines from

    `<model>XF35mmF2 R WR</model>`
    `<model lang="en">XF 35mm f/2 R WR</model>`

to

    `<model>XC35mmF2</model>`
    `<model lang="en">XC 35mm f/2</model>`

Using this updated `mil-fujifilm.xml` both lenses are correctly automatically detected in darktable.

sarunasb:
> IMO, since those are two different lenses (even if optically same), each should have its own `<lens>` element.
> 
> Then, assuming ‘XC35mmF2” and ‘XF35mmF2 R WR’ are what's reported by exiv2, I suggest `<model>` line for each might be:
> 
> `<model>XC 35mm f/2</model>`
> 
> `<model>XF 35mm f/2` R WR
> 
> And then foregoing the `<model lang=...>` lines, since they don't add anything.

It looks like this would be against of what has been used traditionally used in `mil-fujifim.xml`, i.e. short `<model>` line with strictly what is reported by exiv2 and an expanded `<model lang=...>`.

sarunasb:
Pushed to Lensfun in e0ad927.

Issue #1608, opened 2022-02-09
Calibration data for Tamron 28-200 f/2.8-5.6 and Sony 50mm f/1.8 (vignetting added)


ecnivo:
From my file `mil-tamron.xml`:

```kml

    <lens>
        <maker>Tamron</maker>
        <model>E 28-200mm F2.8-5.6 A071</model>
        <mount>Sony E</mount>
        <cropfactor>1</cropfactor>
        <calibration>
            <!-- Taken with ILCE-7M2 Alpha 7 II -->
            <distortion model="ptlens" focal="28" a="0.00611142" b="-0.0180555" c="0.00453851"/>
            <distortion model="ptlens" focal="35" a="0.00655527" b="-0.00664331" c="0.00522778"/>
            <distortion model="ptlens" focal="50" a="0.00304066" b="0.0097271" c="0.00210334"/>
            <distortion model="ptlens" focal="101" a="-0.00172726" b="0.0207933" c="-0.00734779"/>
            <distortion model="ptlens" focal="191" a="0.00959217" b="-0.0226824" c="0.0320856"/>
            <vignetting model="pa" focal="28" aperture="2.8" distance="10" k1="-1.4931" k2="1.2641" k3="-0.5899"/>
            <vignetting model="pa" focal="28" aperture="2.8" distance="1000" k1="-1.4931" k2="1.2641" k3="-0.5899"/>
            <vignetting model="pa" focal="28" aperture="3.2" distance="10" k1="-1.5470" k2="1.4891" k3="-0.7347"/>
            <vignetting model="pa" focal="28" aperture="3.2" distance="1000" k1="-1.5470" k2="1.4891" k3="-0.7347"/>
            <vignetting model="pa" focal="28" aperture="3.5" distance="10" k1="-1.1973" k2="0.7492" k3="-0.2896"/>
            <vignetting model="pa" focal="28" aperture="3.5" distance="1000" k1="-1.1973" k2="0.7492" k3="-0.2896"/>
            <vignetting model="pa" focal="28" aperture="4" distance="10" k1="-0.9020" k2="0.1736" k3="0.0402"/>
            <vignetting model="pa" focal="28" aperture="4" distance="1000" k1="-0.9020" k2="0.1736" k3="0.0402"/>
            <vignetting model="pa" focal="28" aperture="16" distance="10" k1="-0.7969" k2="0.4665" k3="-0.1427"/>
[...]

PR #1609, opened 2022-02-09
Update mil-sony.xml

tuxfanx:
Sony FE 50mm f/1.8: added vignetting correction from issue #1608


tuxfanx (commented):
tested with darktable, looks good

PR #1625, opened 2022-02-13
Added fp and fp L models

qumuq-til:


tuxfanx:
I think the two cameras are mirrorless and should therefore be in the mil file and not in the slr file. If you're ok with that, change your PR so I can merge it.
Have you already checked the automatic detection with Darktable or Rawtherapee?

PR #1629, opened 2022-02-14
Added fp and fp L models

qumuq-til:


qumuq-til:
Not sure if it's possible to just change an existing PR, so I submitted a new one against the correct file. I haven't checked with Darktable et al, but I'm working with SilkyPix that uses lensfun database and since lensfun doesn't have a description for fp series cameras I cannot apply any automatic lens corrections there. If there are any other checks in Darktable and elsewhere for those cameras specifically, let me know.

Issue #1636, opened 2022-02-16
Canon EF 100-200mm f/4.5A missing tca parts


PeterWem:
https://github.com/lensfun/lensfun/issues/1601 missed some focus lenghts for tca between 180-200mm. They are here:

[slr-canon-100-200a.zip](https://github.com/lensfun/lensfun/files/8080057/slr-canon-100-200a.zip)

Also the crop factor for 6D is 1.005

PR #1637, opened 2022-02-16
Updated slr-canon missing tca for 100-200a

PeterWem:
Missing tca 180-200mm. This PR fixes that.
https://github.com/lensfun/lensfun/issues/1636

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -3591,7 +3591,7 @@
         <maker>Canon</maker>
         <model>Canon EF 100-200mm f/4.5A</model>
         <mount>Canon EF</mount>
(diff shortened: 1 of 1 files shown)


sarunasb:
Rationale for changing cropfactor to 1.005?

tuxfanx:
Please commit cleaner PR's or send your changes in an issue. So the calibrators would rather like to deal with your PR's. 
Does a crop factor of 1.005 really make a visible difference to a factor of 1.0?

PeterWem:
It's a 6D. The crop factor is 1.005, check slr-canon.xml for other samples from 6D.
The issue is here https://github.com/lensfun/lensfun/issues/1636

PR #1662, opened 2022-02-22
Add Github action to update lenslist.md on every push

sarunasb:
Github workflow lenslist.yaml is invoked on every push to run `tools/lenslist/show_lensfun_coverage.py`, which produces `lenslist.md`. `show_lensfun_coverage.py` modified lightly for Markdown flavor.


sarunasb:
I'm not sure what would be a good place for lenslist.md in the tree. For now it's in root. Also I don't know how would it be possible to get it to `lensfun.github.io`.

Sturmflut:
@sarunasb Here is an example that uses a Personal Access Token that is configured as a secret in the project settings:

https://github.community/t/push-to-another-repository/205352/7

seebk:
Thanks @sarunasb, that's very good progress to get a self-udpating lens list for the website. I'd prefer to clean up the commit list of this PR and not push the lenslist.md to the source repo but directly to the website repo. In the website repo the lens lists are in the _posts/ subfolder in a separate category. Does not look like it's extensible to what we want now. Not sure why I once decided to solve it like that. Have to refresh my memories a bit, hope I can add your scripts in the next couple of days.

sarunasb:
Yes, it's quite a mess. I didn't find a way to rebase in the web interface.

In any case, it was more of a of Github workflow/action test, and in case having an always up-to-date lenslist in main repo would be useful as well. As it looks like pushing to website repo is possible, then I'll try that in a modified workflow and then I'll make a new pull request.

How about pushing something like current_lenslist.md to _posts/lenslist/? And then have it linked in place of a current link to "develop"?

PR #1673, opened 2022-02-26
Use official lens names for the Canon RF 100-400mm, EF 85 1.4 and EF 28 2.8

piratenpanda:


tuxfanx:
Wouldn't it make things easier if the 1st model element contained exactly what is in the exif data? In the 2nd model parameter, an "official" description can be inserted, if necessary, with `<model lang="en">Canon RF 100-400mm F5.6-8 IS USM</model>"`, for example. 
By the way, if not necessary I would still omit the "Canon" there because it is already described by the maker parameter and would be listed twice in Darktable, see below. I don't know how it behaves in other Software.

I understand the 2. model element more as a type designation, as it also distinguishes Canon, [for example](https://www.canon-europe.com/lenses/lens-type/) and how it is used mostly in the [mil-fujifilm.xml](https://github.com/lensfun/lensfun/blob/master/data/db/mil-fujifilm.xml). The addition of the manufacturer in front of the type designation for me is a sales or advertising description. I would use e.g. "Canon" only if the maker or the first model element does not explain it sufficiently.
This is only my point of view and I know in the database, especially in the Canon xml-files, it is often handled in differently ways. Perhaps we can agree on the use of the 1st model element ;-)

Screenshot_20220226_224137

Screenshot_20220226_214241

Screenshot_20220226_215104

piratenpanda:
See the discussion here https://github.com/lensfun/lensfun/pull/1442 This is because exiv2 had the wrong names. Exiftool is reporting the correct names.

piratenpanda:
I must have remembered incorrectly. This is just about the stuff after "Canon". But still, with "Canon" is the right name and works perfectly fine in darktable.

piratenpanda:
lensnames

And either we change like literally every other lens, also Sigma lenses, Tokina, Tamron, etc. or just the two. Which solution I'd prefer is clear.

tuxfanx:
Just because it's always been done that way doesn't make it correct. For Canon lenses, we can also continue to call it that for the sake of convenience.
I would like to understand it correctly and I am wondering who determines the official model designation and where can I find it out? I think I'm not alone there, see below. Canon itself engraves the label "Canon Lens RF 100-400mm F5.6-8 IS USM" on the lens which tells me it is a lens from the manufacturer Canon with the label RF 100-400mm F5.6-8 IS USM. 

Screenshot_20220227_095722

piratenpanda:
As was discussed here https://github.com/Exiv2/exiv2/pull/1729#issuecomment-865344178 RF lenses should be identified by "RF Lens Type" which is, like it or not, with the Canon in the name. It's not me who made it that way, nor am I introducing new concepts in this PR so I really don't understand why I have to justify those changes here. 

e.g. for the 100-500
rf lens type

tuxfanx:
I didn't ask for a justification. What I wanted was a discussion about how it would be fundamentally correct.
Nevertheless, thanks for your explanations. I think I got it.

piratenpanda:
I'd also love to do it correct and I wouldn't be opposed to groundbreaking changes if it's more correct afterwards. But I tried to follow the previously discussed stuff to keep the changes as minimal as possible for this PR

PR #1809, opened 2022-05-02
Add windows build and tests

payano:


payano:
Do we still want appveyor? Otherwise it would be good to turn it off.
It's also possible to run the build and tests on macos.

payano:
Do you have access to remove appveyor ci/cd?

sarunasb:
20 min. build time for Windows vs. 1min in Ubuntu... Is there anything that could be improved for Windows build in this regard?

Most of the commits to Lensfun are calibration data (changes to XML database). These should not require test builds of libraries, only XML validity check. Not sure if such differentiation between commits would be possible at all, unless data/db is split into git submodule.

payano:
In Linux the glib2 library is already installed in the system, in Windows, we need to build glib2 with dependencies from source and that takes some time. 
If we remove the dependency towards glib it will go much faster, perhaps that's the next step?

we are using glib here:
<pre>
rep -nr "glib" --include="*.c*" --include="*.h" | grep -vE "\.cmake"
tests/test_modifier_coord_scale.cpp:1:#include <glib.h>
tests/test_modifier_coord_scale_old.cpp:1:#include <glib.h>
tests/test_modifier_subpix_old.cpp:1:#include <glib.h>
tests/test_database.cpp:1:#include <glib.h>
tests/test_modifier_coord_distortion.cpp:1:#include <glib.h>
tests/test_modifier.cpp:1:#include <glib.h>
tests/test_modifier_old.cpp:1:#include <glib.h>
tests/test_modifier_color.cpp:1:#include <glib.h>
tests/test_database_old.cpp:1:#include <glib.h>
tests/test_modifier_coord_distortion_old.cpp:1:#include <glib.h>
tests/test_modifier_subpix.cpp:1:#include <glib.h>
tests/test_lens.cpp:1:#include <glib.h>
tests/test_modifier_color_old.cpp:1:#include <glib.h>
tests/test_modifier_coord_geometry_old.cpp:1:#include <glib.h>
tests/test_modifier_coord_geometry.cpp:1:#include <glib.h>
apps/lenstool/lenstool.cpp:6:#include <glib.h>
libs/lensfun/lensfunprv.h:9:#include <glib.h>
libs/lensfun/database.cpp:13:#include <glib/gstdio.h>
docs/example/example.c:8:#include <glib.h>
docs/example/tfun.cpp:6:#include <glib.h>
</pre>

sarunasb:
If Windows test build without glib still makes sense, then yes.

payano:
I think that we should run the tests on the windows target as well. Otherwise we only test that we can build the library.

To improve the build times, perhaps we should investigate google test. I have a Github project which downloads and builds google test and it goes much faster, sub 30 seconds to build google test and the application:
Here is my CMakeLists.txt file:
https://github.com/payano/mMesh/blob/master/test/CMakeLists.txt

If we agree that we wan to rewrite the test cases to use google test instead, I can give it a try.

payano:
time make -j8:

real	0m7,657s
user	0m7,661s
sys	0m0,899s

sarunasb:
@payano Sounds good to me. Please give it a try, if you can.

sarunasb:
Merging this to remove failing Appveyor. If Windows tests can be sped-up or tests replaced with better/efficient, then it can be done in a separate step.

payano:
@sarunasb: I have started to move towards removal of glib. The code base has a lot of dependencies towards glib, I wonder if we should bump the version after that has been done, because I some code needs to be refactored, what do you think?

payano:
By the way, can we remove the builds from appveyor, do you have the login to that site? Otherwise all builds will be marked as failed.
"Some checks were not successful
2 successful and 1 failing checks"

sarunasb:
> By the way, can we remove the builds from appveyor, do you have the login to that site?

I don't have access.

seebk:
I have today uninstalled the Travis and Appveyor support and deleted the connected Appveyor account. However, the current open pull requests still show that AppVeyor build failed. Maybe that's not updated for existing PRs? Lets see...

PR #1827, opened 2022-05-17
added Sony Alpha 7R IIIa

davidmarcoes:
Configuration for the Sony Alpha 7R IIIa. This is the 2021 edition of the 7R III launched in 2018. New model devices require this config for the camera to be properly detected.

More info here: https://www.sony.co.uk/electronics/interchangeable-lens-cameras/ilce-7rm3a

### DIFF
--- a/data/db/mil-sony.xml
+++ b/data/db/mil-sony.xml
@@ -151,6 +151,14 @@
         <mount>Sony E</mount>
         <cropfactor>1</cropfactor>
     </camera>
+    
+    <camera>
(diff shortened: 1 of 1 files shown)


sarunasb (changes requested):
Checking vendors' catalogs and reviews, tt looks like 'IIIA' is preferred over 'IIIa' in camera name. Can you adjust?

davidmarcoes:
Thanks for the feedback @sarunasb, fixed on last commit.

Issue #1472, opened 2021-08-16
lensfun_convert_lcp failed


mikae1:
I'm on Manjaro running `0.3.95`. No `lensfun` dir was present in `~/.local/share/`. I ran `lensfun-update-data` and the dir was created. `~/.local/share/lensfun/updates/version_2/` is now populated with files. 

I then took the LensProfiles from the latest Adobe DNG Converter and ran `lensfun-convert-lcp --prefer-lcp`. That gave me "`Warning: No Lensfun database found.`" followed by "`Warning: Fall back to default mounts for [INSERT LENS NAME].`" for every lcp file in the dir.

What am I missing. Is there still no database? If so, how do I create it?

payano:
@mikae1 : can you test my patch?

PR #1830, opened 2022-05-18
Issue #1472: Add version_ paths to python script

payano:


payano:
Maybe there is a better way to do it?

PR #1832, opened 2022-05-23
Added Pentax K-3 Mark III

HamburgerJungeJr:
Added the configuration for the Pentax K-3 Mark III launched in 2021. 

Details: https://www.ricoh-imaging.eu/de_de/pentax-k-3-mark-iii

I calculated the cropfactor by using the sensor dimensions from the datasheet `(√(24^2+36^2))÷(√(23,3^2+15,5^2))=1,546`

The naming was taken from the EXIF-Data of an image.

### DIFF
--- a/data/db/slr-pentax.xml
+++ b/data/db/slr-pentax.xml
@@ -275,7 +275,7 @@
         <maker>Ricoh Imaging Company, Ltd.</maker>
         <maker lang="en">Pentax</maker>
         <model>Pentax K-3 Mark III</model>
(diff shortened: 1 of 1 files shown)


sarunasb (changes requested):
I think it should be just `<model lang="en">K-3 III</model>`, for consistency among Pentax models in the database.

HamburgerJungeJr:
I just chaged the display name of the camera. 
I thought it should be the official name of the camera but consistency is also nice.

sarunasb (approved):

PR #1844, opened 2022-06-24
Updated slr-canon.xml, Canon EF 35-105mm f/3.5-4.5

PeterWem:
Canon EF 35-105mm f/3.5-4.5 updated


tuxfanx:
Unfortunately I can't check your PR because I don't have any raw images available. If you [upload some raw pics here](https://wilson.bronger.org/calibration) I could check and merge your PR.

PeterWem:
The correction files are 10-12GB. I will send you some sample files. This lens has no focus distance registered, but I created that manually for vignetting as you can see in the PR. The lens has major focus breathing and the vignetting is large at close focus.

PR #1855, opened 2022-07-11
Add distortion correction for TTArtisan APS-C 23mm F1.4.

stativ:
This is a fully manual lens with no EXIF data transmission.

I listed all mounts in which the lens is available, but I don't know whether it's correct given the crop factor differs among the mounts.

Also, I was not sure about the order in which the lenses should be added to misc.xml, so I added it after the last Venus Laowa lens as TTartisan is yet another Chinese brand.

### DIFF
--- a/data/db/misc.xml
+++ b/data/db/misc.xml
@@ -850,8 +850,7 @@
         <mount>Leica L</mount>
         <focal value="23"/>
         <aperture min="1.4" max="16"/>
-        <type>rectilinear</type>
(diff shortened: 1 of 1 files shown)


tuxfanx:
Checked with some raw pictures from the internet. [Uploading](https://wilson.bronger.org/calibration) a few test images makes it easier to check the calibration. 
The sequence of the entries is basically irrelevant. Most often, the data for new lenses will be added at the end of the file.

PR #1859, opened 2022-07-17
Fix out-of-bound memory write in `lf_mlstr_add`

### LINKED ISSUE #1858: Segfault when loading a lens with a duplicate model field
While doing some testing for https://github.com/lensfun/lensfun/issues/1846, I encountered a segfault in lensfun (ran through darktable). The segfault was triggered when I mistakenly added a duplicate `<model>` field to a lens, like so:
```xml
    <lens>
        <maker>Sigma</maker>
        <model>56mm F1.4 DC DN | Contemporary 018</model>
        <model lang="en">Sigma 56 mm F1.4 DC DN | C 018</model>
        <model>SIGMA 56mm F1.4 DC DN | Contemporary 018</model>
```

I hunted down the error to the `lf_mlstr_add()` function, which does not handle this scenario correctly. There is code in this function to handle the replacement of the "default" model name, when it is duplicated like in the example above. However, the code only works if the new name is shorter than the previous name, or if there is no language-specific entry already in the string. If it is longer, and there is a language-specific entry, then the code causes out-of-bound memory write, which then unleashes demons and other vulnerability niceties.

I have fixed the code and will open a pull request shortly.

For reference, running darktable from the terminal, I got the following cryptic error
```
corrupted size vs. prev_size
Aborted (core dumped)
```

cschreib:
Closes #1858.

### DIFF
--- a/libs/lensfun/auxfun.cpp
+++ b/libs/lensfun/auxfun.cpp
@@ -98,9 +98,21 @@ LF_EXPORT lfMLstr lf_mlstr_add (lfMLstr str, const char *lang, const char *trstr
         /* Replace the default string */
         size_t def_str_len = str ? strlen (str) + 1 : 0;
 
-        memcpy (str + trstr_len, str + def_str_len, str_len - def_str_len);
-
-        str = (char *)g_realloc (str, str_len - def_str_len + trstr_len + 1);
+        if (str)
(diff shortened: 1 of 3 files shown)


sarunasb:
Thank you for contributing.

PR #1865, opened 2022-07-27
add distortion correction for Sigma 28-70mm F2.8 DG DN | C 021

### LINKED ISSUE #1848: Calibration upload 0e4a10
Calibration images were uploaded to the directory that starts with “`0e4a10_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> bonjour
> pour étalonnage 
> objectif Sigma 28-70 DG DN
> 
> merci
> andré

tuxfanx:
fixes #1848
Unfortunately the automatic lens detection in darktable does not work with panasonic cameras and this lens at the moment. With Sony cameras, the automatic detection works correctly.
@sarunasb: do you have any ideas what the reason for that could be? Is the cause in exiv2 or darktable? [2 pictures more](https://jonani1.my-router.de/nextcloud/index.php/s/e7pQPnSZoS6LLnH)

### DIFF
--- a/data/db/mil-sigma.xml
+++ b/data/db/mil-sigma.xml
@@ -695,4 +695,21 @@
         </calibration>
     </lens>
 
+    <lens>
+        <maker>Sigma</maker>
+        <model>28-70mm F2.8 DG DN | Contemporary 021</model>
+        <mount>Sony E</mount>
+        <mount>Leica L</mount>
+        <cropfactor>1.0</cropfactor>
+        <calibration>
+            <!-- Taken with Panasonic Lumix DC-S5 -->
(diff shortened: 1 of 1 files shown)


sarunasb:
Cameras' firmware perhaps are putting slightly different lens identification into EXIF.

Also, do we know whether images are taken with the same lens? There might be newer/older versions of lens or its firmware.

Given two different strings for the same lens, the longer, fuller string in `model` will have a better chance for lens being recognised.

PR #1916, opened 2022-10-18
add cameras, changes and a lens to rf-leica.xml

tuxfanx:
added Summicron-M 1:2/28 ASPH.
added M9, M10, M10-P, M10-R, M11
changed maker from "Leica" to "Leica Camera AG"
corrected model name for Elmarit-M 1:2.8/28 ASPH.

### DIFF
--- a/data/db/rf-leica.xml
+++ b/data/db/rf-leica.xml
@@ -19,6 +19,13 @@
 
     <camera>
         <maker>Leica Camera AG</maker>
+        <model>Leica M Monochrom (Typ 246)</model>
+        <mount>Leica M</mount>
+        <cropfactor>1</cropfactor>
+    </camera>
+
+    <camera>
+        <maker>Leica Camera AG</maker>
         <model>M9 Digital Camera</model>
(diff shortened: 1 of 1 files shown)


data/db/rf-leica.xml
```diff
@@ -11,15 +11,50 @@
     </mount>
 
     <camera>
-        <maker>Leica</maker>
+        <maker>Leica Camera AG</maker>
```
kmilos, line 14: Should probably keep <maker lang="en">Leica</maker> as well - both libraw and rawspeed sanitize the maker to just "Leica" (for RWLs mostly), which could be passed to lensfun.

kmilos: You're probably right that it should be safe to stick w/ the original Exif reported value - darktable passes that one indeed for lensfun lookup, and not the rawspeed sanitized short name used elsewhere in the UI, so I was maybe too quick to comment. ;) Probably fine to proceed like this, and wait for any regression reports... (hopefully none!)

Issue #237, opened 2017-08-21
Calibration upload b035fd


bronger:
Calibration images were uploaded to the directory that starts with “`b035fd_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Meike mk-f 28mm f/2.8 Fuji X mount

ludoo:
I did the calibration workflow for barrel distortion, and came up with the pasted XML below. Screenshots from Darktable show it seems to be working. It's my first one, so it could probably be done better. ;)

<lensdatabase version="1">
    <mount>
        <name>Meike</name>
        <compat>Fujifilm X</compat>
    </mount>
    <lens>
        <maker>Meike</maker>
        <model>MK-F 28mm f/2.8</model>
        <model lang="en">MK-F 28mm f/2.8</model>
        <mount>Fujifilm X</mount>
        <cropfactor>1.529</cropfactor>
        <calibration>
            
            <distortion model="ptlens" focal="28" a="0.0036057" b="-0.0218902" c="0.003422" />
        </calibration>
    </lens>
</lensdatabase>

![screenshot from 2017-08-21 20-17-22](https://user-images.githubusercontent.com/66007/29532875-9a4aac2c-86af-11e7-9be0-930c408a91dd.png)
![screenshot from 2017-08-21 20-18-12](https://user-images.githubusercontent.com/66007/29532876-9a4cd24a-86af-11e7-8704-5ef5284cb524.png)

ludoo:
aaand XML did not survive the comment formatter, here is the relevant line

distortion model="ptlens" focal="28" a="0.0036057" b="-0.0218902" c="0.003422"

bronger:
@uploader I could only measure distortion.  The image was too soft for TCA measurement.

PR #1964, opened 2023-01-09
Make Meike 28mm f/2.8 lens name consistent with other Meike's names

alexmalykh:
Hello there. Please consider this little fix for non-exif Meike lens.

### DIFF
--- a/data/db/misc.xml
+++ b/data/db/misc.xml
@@ -743,7 +743,7 @@
 
     <lens>
         <maker>Meike</maker>
-        <model>Meike MK-F 28mm f/2.8</model>
+        <model>Meike 28mm f/2.8</model>
         <mount>Fujifilm X</mount>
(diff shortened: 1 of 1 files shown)


sarunasb:
Meike MK-F 28mm added from calibration upload https://github.com/lensfun/lensfun/issues/237. 
@alexmalykh Not sure what MK-F indicates, can't find it anywhere else, but it in #237 upload. Do you mean that lens is purely mechanica, does not provide any metadata in image files?

alexmalykh:
Yes, it does not provide any metadata, it's a purely manual mechanical manual focus lens. There are at least two other Meike lenses from this family: 25mm f/1.8 and 35mm f/1.7 (the latter I own too) - but there are no any extra markings like "MK-F" neither in the database nor on the lenses' label rings for those ones.

PR #1969, opened 2023-01-14
Added support for Olympus E-M10 Mark III S

dorpauli:
When editing my RAW photos of the Olympus E-M10 Mark III S in Darktable, I always had to manually set the correct camera model. But in principle it is the E-M10 Mark III with silent shutter. 

Now I was tired of it and found a solution. If the multiple use of "model" is not desired, one can of course still add a separate "camera" entry. So it has now worked for me, however!

### DIFF
--- a/data/db/mil-olympus.xml
+++ b/data/db/mil-olympus.xml
@@ -199,6 +199,7 @@
         <maker>Olympus Corporation</maker>
         <maker lang="en">Olympus</maker>
         <model>E-M10 Mark III</model>
+        <model>E-M10MarkIIIS</model>
         <model lang="en">E-M10 III</model>
         <mount>Micro 4/3 System</mount>
         <cropfactor>2</cropfactor>


tuxfanx:
Your reqest did not work with "E-M10 III" raw files. I added an "S" to the model name of "E-M10 III" (model name is "E-M10 III S" now) what should work for both cameras. Please test it with your pictures.

dorpauli:
> Your reqest did not work with "E-M10 III" raw files. I added an "S" to the model name of "E-M10 III" (model name is "E-M10 III S" now) that should work for both cameras. Please test it with your "E-M10 III S" pictures.

No it is not working. If i remove spaces in line 201 (E-M10MarkIIIS) it is!

tuxfanx:
Thanks, i think we need to create a separate camera entry because without the spaces it will not work for the E-M10 III.
Unfortunately, I could not find a raw image on the internet that was taken with the S version, I only had a jpg for the S version and a [raw image of the E-M10 III](https://www.dpreview.com/sample-galleries/6825955048/olympus-om-d-e-m10-iii-gallery) to test what worked for me in Darktable for both pictures. Please could you [upload a raw image from your camera here](https://wilson.bronger.org/calibration) for testing?

sarunasb:
Yes, separate `<camera>` element is of course necessary.

Spaces should not make a difference for matching and auto detection, so there might be something else.

RAW samples on raw.pixls.us?

tuxfanx:
@sarunasb: thanks for the hint!
@dorpauli: I found a raw image on pixls.us and created a separate camera profile.

PR #1973, opened 2023-01-24
Fix argument -c for dcraw_emu

twbaumann:
I tried to run the current new version of calibrate.py for the calibration images of my Fujifilm G lenses. It uses dcraw_emu instead of dcraw. The program exited with the error message 'Non-numeric argument to "-c"' which is caused by the fact that '-c' means different things to dcraw and dcraw_emu:

```
dcraw:     -c            Write decoded images or thumbnails to standard output.
dcraw_emu: -c float-num  Set adjust maximum threshold (default 0.75)
```

The correct emulation of the argument '-c' would be '-Z -', so I changed "-c" to ["-Z", "-"] for output on STDOUT. Additionally, for TIFF file output I added ["-Z", "tiff"] to the dcraw_options because otherwise dcraw_emu would append '.tiff' to the raw file extension (\<basename\>.\<raw\>.tiff) and the file would not be found during further execution, leading to another error message.

This fix works for me on MacOS 10.14.6 with Python 3.10 and dcraw_emu being provided by MacPorts as libraw @0.21.1_0. I tested all three of distortion, TCA and vignetting.

### DIFF
--- a/tools/calibrate/calibrate.py
+++ b/tools/calibrate/calibrate.py
@@ -117,7 +117,7 @@ def generate_raw_conversion_call(filename, dcraw_options):
         result = [convert_program, filename]
         if "-4" in dcraw_options:
             result.extend(["-colorspace", "RGB", "-depth", "16"])
-        result.append("tiff:-" if "-c" in dcraw_options else basename + ".tiff")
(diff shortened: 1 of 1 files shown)


sarunasb:
@twbaumann Thank you for fixing calibrate.py.

PR #1987, opened 2023-02-17
use threadsafe uselocal (_configthreadlocale for Windows)

### LINKED ISSUE #1984: Interference between lensfun and gphoto causes darktable to crash
The original bug submission started with darktable.  darktable completes an import of images, it begins to process thumbnails.  During that process, darktable sends a request to lensfun.  At that time, darktable crashes.  Crash is caused by lensfun.  See below.

Below is a link to the crash submission at darktable.  It includes a full crash report.  
https://github.com/darktable-org/darktable/issues/13637#issue-1583565327

Github member parafin reviewed the crash report, and matched it to the problematic code.  I'm not a programmer, so I quote his comments directly below ...

parafin wrote "The crash seems to be caused by lensfun doing setlocale call in thread 6 while gphoto is doing sprintf in thread 11 (which depends on locale). So it's an interference between two darktable dependencies (lensfun and gphoto) due to multi-threading. This has to be fixed upstream in lensfun, it mustn't do setlocale calls, since it's not thread-safe and affects the whole application. 

Offending line is here: https://github.com/lensfun/lensfun/blob/release_0.3.x/libs/lensfun/lens.cpp#L253

I suggest to report it to lensfun."

I hope this submission is helpful.

jonas-ott:
fixes #1984

Instead of `setlocale()` we can use [uselocal](https://man7.org/linux/man-pages/man3/uselocale.3.html) or [_configthreadlocale](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/configthreadlocale?view=msvc-170) for Windows. Both change the locale only for the lensfun thread.
I did not test if the crash in darktable is fixed but if this `setlocale()` is the issue then it should be.

This should probably be part of the 0.3.4 release.

### DIFF
--- a/libs/lensfun/database.cpp
+++ b/libs/lensfun/database.cpp
@@ -738,9 +738,14 @@ lfError lfDatabase::Load (const char *errcontext, const char *data, size_t data_
     };
 
     /* Temporarily drop numeric format to "C" */
-    char *old_numeric = setlocale (LC_NUMERIC, NULL);
-    old_numeric = strdup(old_numeric);
-    setlocale(LC_NUMERIC,"C");
+#if defined(_MSC_VER)
+    _configthreadlocale(_ENABLE_PER_THREAD_LOCALE);
+    setlocale (LC_NUMERIC, "C");
+#else
+    auto loc = uselocale((locale_t) 0); // get current local
+    auto nloc = newlocale(LC_NUMERIC_MASK, "C", (locale_t) 0);
+    uselocale(nloc);
+#endif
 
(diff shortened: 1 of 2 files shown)


sarunasb:
Similar locale manipulations are done also in `libs/lensfun/database.cpp`. Should those be changed as well?

jonas-ott:
Yes, probably.

I did the changes in in `libs/lensfun/database.cpp`  and some tests failed. I figured out that restoring with `uselocale(LC_GLOBAL_LOCALE);` isn't sufficient and it seems necessary to store the old locale. I'm not sure why but let's hope it's correct this time.

PR #2004, opened 2023-03-10
Add vignetting for Samyang 12mm AF

interru:


tuxfanx:
I would like to add your data to the db, but personally I only merge calibration data when I have been able to test it on some images.
Unfortunately I could not find any raw images for this lens. So it would be nice if you could [upload some pictures here](https://wilson.bronger.org/calibration). 3 normal shots and 3 shots you used for calibration would be enough.

luzpaz:
@interru :point_up_2:

luzpaz:
> @interru 👆

PR #1999, opened 2023-03-02
Fix checking Windows platform for non-msvc toolchain

Biswa96:
Fixes https://github.com/lensfun/lensfun/issues/1998

Issue #1998, opened 2023-03-02
MSYS2/MinGW, GCC: compilation errors related to locale_t type


LigH-de:
Trying to compile ffmpeg with lensfun in the [media-autobuild suite](https://github.com/m-ab-s/media-autobuild_suite), an MSYS2/MinGW environment which keeps itself up to date before compiling (here with GCC 12) many modules linked into ffmpeg, verbosely logging all issues.

Excerpt:
```
...
G:/MABS/build/lensfun-git/libs/lensfun/database.cpp: In member function 'lfError lfDatabase::Load(const char*, const char*, size_t)':
G:/MABS/build/lensfun-git/libs/lensfun/database.cpp:826:27: error: 'locale_t' was not declared in this scope; did you mean '_locale_t'?
  826 |     auto loc = uselocale((locale_t) 0); // get current local
      |                           ^~~~~~~~
      |                           _locale_t
G:/MABS/build/lensfun-git/libs/lensfun/database.cpp:826:16: error: 'uselocale' was not declared in this scope; did you mean 'setlocale'?
  826 |     auto loc = uselocale((locale_t) 0); // get current local
      |                ^~~~~~~~~
      |                setlocale
G:/MABS/build/lensfun-git/libs/lensfun/database.cpp:827:27: error: 'LC_NUMERIC_MASK' was not declared in this scope; did you mean 'LC_NUMERIC'?
  827 |     auto nloc = newlocale(LC_NUMERIC_MASK, "C", (locale_t) 0);
      |                           ^~~~~~~~~~~~~~~
      |                           LC_NUMERIC
G:/MABS/build/lensfun-git/libs/lensfun/database.cpp:827:59: error: expected ')' before numeric constant
  827 |     auto nloc = newlocale(LC_NUMERIC_MASK, "C", (locale_t) 0);
      |                          ~                                ^~
      |                                                           )
G:/MABS/build/lensfun-git/libs/lensfun/database.cpp:827:17: error: 'newlocale' was not declared in this scope; did you mean 'setlocale'?
  827 |     auto nloc = newlocale(LC_NUMERIC_MASK, "C", (locale_t) 0);
      |                 ^~~~~~~~~
      |                 setlocale
G:/MABS/build/lensfun-git/libs/lensfun/database.cpp:858:5: error: 'freelocale' was not declared in this scope; did you mean '_free_locale'?
  858 |     freelocale(nloc);
      |     ^~~~~~~~~~
      |     _free_locale
[...]

PR #2011, opened 2023-03-24
GitHub Actions: MSYS2 workflow

jonas-ott:
Windows build with [MSYS2](https://www.msys2.org/) using https://github.com/marketplace/actions/setup-msys2.
In comparison to the CMake-Windows workflow, it uses GCC instead of MSVC and installing glib2 is way faster.

Would have caught  https://github.com/lensfun/lensfun/issues/1998.
On top of https://github.com/lensfun/lensfun/pull/1999, so the build doesn't fail.

### DIFF
--- a/.github/workflows/cmake-msys2.yml
+++ b/.github/workflows/cmake-msys2.yml
@@ -1,0 +1,36 @@
+name: CMake-MSYS2
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  BUILD_TYPE: Release
+
+jobs:
+  msys2-ucrt64:
+    runs-on: windows-latest
+    defaults:
+      run:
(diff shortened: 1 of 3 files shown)


.github/workflows/cmake-msys2.yml
```diff
@@ -0,0 +1,36 @@
...
+        shell: msys2 {0}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: UCRT64
+          update: true
+          pacboy: >-
+            cmake:p
+            gcc:p
```
kmilos, line 26: You can change this to MSYS2 environment agnostic `cc:p` so adding e.g. CLANG64 or CLANGARM64 in the future to the `msystem` matrix would be seamless.

PR #2094, opened 2023-08-12
Stop using INT_MIN/INT_MAX for floats

pinotree:
`INT_MIN`/`INT_MAX` are meant for the `int` type, and thus casting them to a different type (`float`) means loss of precision which may lead to failed comparisons later on.

Instead use the proper limits using `std::numeric_limits`.


pinotree:
Thanks @sarunasb! Since the problem affects also lensfun 0.3, I backported this to the `release_0.3.x` branch: #2098.

PR #2118, opened 2023-10-19
Handle absolute and relative paths for pkgconf generation

kmilos:
Resolves https://github.com/lensfun/lensfun/issues/1938


kmilos:
It seems setuptools needs to be also installed for the Windows job...

PR #2128, opened 2023-11-25
install setuptools in cmake-windows workflow

jonas-ott:
The CMake-Windows workflow currently fails with `ModuleNotFoundError: No module named 'setuptools'`. Fixed by installing the python package setuptools. Implemented as described [here](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#installing-dependencies).


kmilos:
[Indeed](https://github.com/lensfun/lensfun/pull/2118#issuecomment-1770879997).

PR #2130, opened 2023-12-05
fix year 2038 problem

jonas-ott:
`ReadTimstamp()` returns a `long int`. Converted to an `int` it will overflow in 2038.


libs/lensfun/database.cpp
```diff
@@ -94,11 +94,11 @@ lfError lfDatabase::Load ()
 {
   lfError err = LF_NO_ERROR;
 
-    const int timestamp_system =
+    const auto timestamp_system =
```
payano, line 97: looks good. I like beeing explicit about the data types but this will work fine :)

CoolCat467 (approved):
Personally, I think auto might be a better idea because what if the type changes again in new versions? Using auto will avoid needing another change if anything happens in the future.

libs/lensfun/database.cpp
sarunasb: I assume the change was tested on a system with date set past 2038?

payano:
> Personally, I think auto might be a better idea because what if the type changes again in new versions? Using auto will avoid needing another change if anything happens in the future.

I think it's better to be explicit about the data types you are declaring. And if the implementation you are using have changed the return type, this is a good way to be notified about that and make the correct changes in order to keep the behaviour the same as before.

libs/lensfun/database.cpp
jonas-ott: > I assume the change was tested on a system with date set past 2038?

Yes, i did some basic testing.
The system date doesn't matter, it's about the value in in the `timstamp.txt` file in the database. With the change a timestamp after 2038 is no longer interpreted as a negative number.

jonas-ott:
> And if the implementation you are using have changed the return type, this is a good way to be notified about that

Only if we would activate `Wconversion` warnings, but that would generate a lot of warnings.
  
> It is also good to be explicit from a code reviewer standpoint.

You would still have to check if the data type matches the function, which didn't happen when this code was written. With auto this bug wouldn't have happened.

But yeah, it's also a personal preference and not that important.

payano:
> > And if the implementation you are using have changed the return type, this is a good way to be notified about that
> 
> Only if we would activate `Wconversion` warnings, but that would generate a lot of warnings.
> 
> > It is also good to be explicit from a code reviewer standpoint.
> 
> You would still have to check if the data type matches the function, which didn't happen when this code was written. With auto this bug wouldn't have happened.
> 
> But yeah, it's also a personal preference and not that important.

Maybe we should add `Wconversion` and start to remove conversions that could cause problems.
It depends on how we treat the "master" branch. I think it should be a develop branch and when we feel ready we create a release, either a new release branch or just a release on the develop branch.

jonas-ott:
> Maybe we should add `Wconversion` and start to remove conversions that could cause problems.
Sure,

jonas-ott:
Sorry, That was the wrong button


> Maybe we should add Wconversion and start to remove conversions that could cause problems.

Sure, but that is work for the future.

libs/lensfun/database.cpp
CAHEK7: Unfortunately it does not fix the problem for MSVC, because `sizeof(long int)` is 4 bytes for this compiler even for 64bit systems and `ReadTimestamp` returns `long int`

CAHEK7 (changes requested):
The whole `ReadTimestamp` function should be fixed to avoid this problem for MSVC compiler.

CAHEK7 (approved):

Issue #2131, opened 2023-12-09
calibration data hosting


jonas-ott:
Currently, `lensfun-update-data` downloads calibration data from `https://lensfun.sourceforge.net/db/` and `https://wilson.bronger.org/lensfun-db/`. I'm not sure how those are hosted and updated, but I believe both are reliant on @bronger.

As the updates are just files hosted on a web server it would be possible to use https://lensfun.github.io for hosting. The site already updates with every new lens (make-lenslist workflow), so just building the update files and putting them in a folder would be easy.

GitHub pages have ["a soft bandwidth limit of 100 GB per month"](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#usage-limits)  equivalent to more than 250,000 updates (minus the normal traffic). I don't have any usage data but it feels like this should be sufficient.  

I have already played a bit with this, and if there are no objections to the idea, I could open a pr.

Edit: I have created three pull requests for doing this.

1. https://github.com/lensfun/lensfun/pull/2132 Preparation: Python script for creating the update files
2. https://github.com/lensfun/lensfun.github.io/pull/9 The workflow for generating the files and deploying the site
3. https://github.com/lensfun/lensfun/pull/2133 updating the sources in lensfun-update-data

bronger:
Good idea! So, a job on GitHub that is triggered after every commit would update the DB files?

jonas-ott:
Yes, so the `make-lenslist` workflow creates a commit in the lensfun.github.io repo which triggers the `pages-build-deployment` workflow. This workflow (default GitHub Pages behavior) can be replaced by a custom workflow that first builds the update files. The files would not be committed to the repo, so it doesn't grow in size.

bronger:
I suggest to replace SourceForge with GitHub. I can continue to operate my own site as a backup, but would not object to letting it fade out over the next two or three years.

PR #2132, opened 2023-12-13
Add tool for generating database updates

jonas-ott:
Preparation for https://github.com/lensfun/lensfun/issues/2131.

`tools/update_database/follow_db_changes.py` currently generates the files for updates and a few other things.
For hosting the files on `lensfun.github.io` we need a tool that just generates the files. Therefore most of the generation and conversion of the files is moved to `tools/update_database/xml_converter.py` and used in `follow_db_changes.py` and the new `tools/update_database/generate_db.py`.

`tools/lensfun_convert_db_v2_to_v1.py`, used for making 0.3.x releases already did most of what `generate_db.py` does, but not everything needed. Now that we have `generate_db.py` it can be removed.

Edit: GitHub thinks `lensfun_convert_db_v2_to_v1.py` was moved to `xml_converter.py`, actually  the code is copied from  `follow_db_changes.py`.


jonas-ott:
Added a change to the make-lenslist workflow. If there is a correction to the data (for example https://github.com/lensfun/lensfun/commit/793d1c4bdb7a1dc381d3ef9af9f53d10d8dd3124) the lenslist doesn't change and the workflow would fail. This would result in hosting the wrong data because the workflow on lensfun/lensfun.github.io wouldn't be triggered.

PR #2133, opened 2023-12-13
replace "sourceforge.net" with "lensfun.github.io" in lensfun-update-data

### LINKED ISSUE #2131: calibration data hosting

jonas-ott:
The final part of https://github.com/lensfun/lensfun/issues/2131 (fixes #2131)

Only merge after https://github.com/lensfun/lensfun.github.io/pull/9!

Replaces the Sourceforge source with "lensfun.github.io" (https://github.com/lensfun/lensfun/issues/2131#issuecomment-1848921025)

### DIFF
--- a/.github/workflows/make-lenslist.yml
+++ b/.github/workflows/make-lenslist.yml
@@ -47,7 +47,9 @@ jobs:
         git add _posts/lenslist/2999-12-31-Lenslist-master.md
         git config --local user.email "github-actions[bot]@users.noreply.github.com"
         git config --local user.name "github-actions[bot]"
-        git commit -m "Update master lens list from GitHub Actions"
+        # A change without adding a new lens or camera doesn't change the lenslist and the commit would fail.
+        # With --allow-empty the commit succedes and the workflow for data hosting on lensfun/lensfun.github.io is triggered.
+        git commit --allow-empty -m "Update master lens list from GitHub Actions"
         git push https://$USERNAME:$REPO_KEY@github.com/lensfun/lensfun.github.io.git
       env:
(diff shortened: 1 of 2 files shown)


jonas-ott:
Added a change to the make-lenslist workflow. If there is a correction to the data (for example https://github.com/lensfun/lensfun/commit/793d1c4bdb7a1dc381d3ef9af9f53d10d8dd3124) the lenslist doesn't change and the workflow would fail. This would result in hosting the wrong data because the workflow on lensfun/lensfun.github.io wouldn't be triggered.

PR #2139, opened 2024-01-03
Add further mounts to Sigma 100-400mm F5-6.3 DG OS HSM

mime01:
The Sigma 100-400mm F5-6.3 DG OS HSM | Contemporary is also available for the "Canon EF", "Sigma SA" and "Nikon F AF" mounts.


tuxfanx:
I can't find the Sony E-mount for this lens in the manufacturer's technical data either. Is there a confusion with the 100-400mm F5-6.3 DG DN OS | Contemporary lens?
However, the lenses do not have the same optical structure.

tuxfanx:
May be a Sigma MC 11 adapter was used and a Sony mount connection was therefore specified. However, this is not the correct approach.

PR #2143, opened 2024-01-05
Vignetting correction for Sigma 18-250mm f/3.5-6.3 DC OS Macro HSM and Sigma 24mm F1.4 DG HSM | A

Macchiato17:
This PR provides vignetting correction for 
- Sigma 18-250mm f/3.5-6.3 DC OS Macro HSM (at 18mm, 32mm, 52mm, 130mm, 250mm)
- Sigma 24mm F1.4 DG HSM | A 
Values were determined using a Nikon D7500, thus valid for a crop factor of 1.5.

### DIFF
--- a/data/db/slr-sigma.xml
+++ b/data/db/slr-sigma.xml
@@ -2152,7 +2152,7 @@
         </calibration>
     </lens>
 
-     <lens>
+    <lens>
         <maker>Sigma</maker>
         <model>Sigma 24mm F1.4 DG HSM | A</model>
         <model lang="en">Sigma 24mm f/1.4 DG HSM | A</model>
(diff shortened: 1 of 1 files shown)


Macchiato17:
I have the impression, that the correction parameters provided herein are too aggressive, since quite a large area of the picture was influenced by the correction. Perhaps this is right, but I'm not sure. Therefore I would like to review the current vignetting parameters by redoing the photos with a different diffuser.

tuxfanx:
@Macchiato17 
An evenly lit area is also important. When artificial light is used on a wall surface, the edges of the image are often slightly weaker than the center of the image.
I would like to test your profiles and add them to the db if the correction is good, but personally I only merge calibration data when I have been able to test it on some raw pictures.
Unfortunately, I could not find any raw images for your lenses. So it would be nice if you could upload some pictures [here](https://wilson.bronger.org/calibration). 2-3 normal shots (even surfaces with little structure with the aperture open are good e.g. plain, smooth walls or a blue sky) and 2-3 shots you used for calibration would be enough.

Macchiato17:
I really appreciate your kind offer to check the calibration data on your side. I uploaded the RAW files that I used for creating the vignetting correction parameter set. I also added some "normal" shots I took with both lenses today. If you need other pictures, just give me a ping.
Regarding vignetting correction, I mainly followed the [guide on pixls.us](https://pixls.us/articles/create-lens-calibration-data-for-lensfun/) again. All photos were taken outside, against a grayish, featureless sky, using a diffuser directly mounted in front of the lens. The camera was mounted on a tripod. Focus set to infinity.

tuxfanx:
Many thanks!
The profile for the 24mm lens is ok. It may have only a minimal overcorrection.
Unfortunately, the images for the zoom lens are missing on the server. Two uploads have been created for you, but without a zip file with images. The same thing happened to someone else. Maybe the server has a problem. Please try once more.
If it does not work, I`ll send you another cloud address for upload.

Macchiato17:
OK, I tried again, please check if the two ZIPs were uploaded 😃

Macchiato17:
Thanks for merging 😄

PR #2148, opened 2024-01-12
Added IQ180, IQ140 and Schneider LS 80mm f/2.8 to 6x6.xml

BAGELGENESIS:


tuxfanx:
I would like to add your data to the db, but personally I only merge calibration data when I have been able to test it on some images.
Unfortunately, I could not find any raw images for the 80mm lens. So it would be nice if you could [upload some pictures here](https://wilson.bronger.org/calibration). 2 normal shots and 2 shots you used for calibration would be enough.

tuxfanx:
Ok, many thanks for the Info!

PR #2150, opened 2024-01-13
Add vignetting data for the Canon RF 14-35

piratenpanda:
Taken with a Gerd Neumann flatfield mask


piratenpanda:
still not completely happy

piratenpanda:
at f4 it's really hard to not have circular residue in the image. 

grafik

Fit looks like this:
grafik

Any ideas?

piratenpanda:
tried again with my 120cm softbox and opal white (milky) acryl glass,  same results. So I suggest to use these values until someone comes up with a better idea
grafik

cytrinox:
@piratenpanda Can you share (all) your flats with me? I've calibrated my RF-14-35 long time ago but it was not perfect, so I've never published it (minimal dark edges at 14mm). But I remember to hit the same ringing effect but it has gone after changing the exposure time for new set of flats.

 I want to compare your flats againts my profile - if it has no ringing I think it's better to use my profile :-)

cytrinox:
FYI: my profile: [rf14_35.zip](https://github.com/lensfun/lensfun/files/13932489/rf14_35.zip)

piratenpanda:
with your profile it's basically a bigger circle and more falloff to the edges:
grafik

This lens is a bit annoying..

piratenpanda:
this is the best I can do:
grafik

with 
`<vignetting model="pa" focal="14.0" aperture="4.0" distance="0.19" k1="-0.5417235" k2="-0.4968774" k3="0.1342000" />`

and a new file
https://www.pandainthecloud.de/nextcloud/index.php/s/t7xCmGYCHCtt2dH

piratenpanda:
This does not get better unfortunately

PR #2151, opened 2024-01-14
Add the Canon RF 135mm F1.8L IS USM

piratenpanda:


tuxfanx:
In the event of possible future problems, entering the camera model used can sometimes be helpful.

PR #2154, opened 2024-01-23
Calibration data for various Phase One XF/Mamiya 645 lenses

BAGELGENESIS:
Added IQ180 and IQ140 digital backs.
Added the following lenses:
- Mamiya 120mm f/4.0
- Mamiya 150mm f/3.5
- Schneider LS 80mm f/2.8

### DIFF
--- a/data/db/6x6.xml
+++ b/data/db/6x6.xml
@@ -104,6 +104,88 @@
     </lens>
 
     <lens>
+        <maker>Mamiya</maker>
+        <model>120mm f/4.0</model>
+        <mount>Mamiya 645</mount>
+        <cropfactor>0.644</cropfactor>
+        <calibration>
+            <distortion model="ptlens" focal="120.0" a="0.000368115" b="-0.00348236" c="0.00443514" />
+            <tca model="poly3" focal="120.0" br="-0.0000119" vr="0.9999574" bb="0.0000050" vb="0.9999868" />
+            <vignetting model="pa" focal="120.0" aperture="4.0" distance="0.402" k1="0.0501982" k2="-1.1374235" k3="0.5416169" />
+            <vignetting model="pa" focal="120.0" aperture="4.0" distance="0.804" k1="-1.3133803" k2="1.1395689" k3="-0.4782986" />
+            <vignetting model="pa" focal="120.0" aperture="4.0" distance="2.412" k1="-1.7653588" k2="1.8503581" k3="-0.8170525" />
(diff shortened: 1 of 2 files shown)


tuxfanx:
I think there is a mistake in the profiles.
The aspect ratio is not specified, [see here](https://wilson.bronger.org/lensfun/el_lens.html).
It should be specified when creating the profile. I don't know whether the missing information in the different applications leads to incorrect corrections. If no information is provided, lensfun assumes an aspect ratio of 3:2.

The model name should include "Mamiya" like the other Mamiya lenses. You can use e.g. `<model lang="en">Mamiya 20mm f/4.0</model>` on an additional line.

It is sometimes helpful if the lens profile contains information about which camera was used e.g. `<!-- Taken with  Phase One IQ 180-->`

tuxfanx:
I found the pictures, thank you very much.
I looked at your changes in the new PR. It is not allowed to use the `<model>` identifier twice. [This identifier](https://wilson.bronger.org/lensfun/el_lens.html) is for the automatic lens detection. Please use the identifier `<model lang="en">Mamiya ....</model>`. This identifier is then only responsible for the output on the screen and the [lensfun coverage website](https://wilson.bronger.org/lensfun_coverage.html).

The line with information about the camera used should be part of the calibration data obtained with it. The calibration data of a lens profile could have been created with different camera models.

Here is an example from the database:
```
<lens>
        <maker>Nikon</maker>
        <model>Nikon AF-S Nikkor 16-35mm f/4G ED VR</model>
        <model lang="en">Nikkor AF-S 16-35mm f/4G ED VR</model>
        <mount>Nikon F AF</mount>
        <cropfactor>1</cropfactor>
        <calibration>
            <!-- Taken with Nikon D750 -->
            <distortion model="ptlens" focal="16" a="0.01589" b="-0.08067" c="0.06712"/>
```

tuxfanx:
In my OS (Debian Bookworm) lensfun does not perform picture correction under Darktable if the aperture range is included in the model name. If I paste the original model name back without the aperture range, the correction is carried out. However, I have to select all the lenses manually. Seems that Exiv2 has problems with the Phase One exif data.
Does the picture correction work on your OS with the aperture range in the model name?

tuxfanx:
Thank you for the hints!
After I converted the images with the Adobe DNG converter, the automatic correction also works in my system. We should leave the model names as they are now. The problem lies with exiv2 and the IIQ file format.
I would still like to change one thing. The profile for the Schneider lens should not be in the slr-schneider.xml. I think it should be in the 6x6.xml. It's a medium format lens, isn't it?

PR #2160, opened 2024-02-02
Add C interface for EnableDistortionCorrection

CAHEK7:
Add C interface for `lfModifier::EnableDistortionCorrection (const lfLensCalibDistortion& lcd)`, addresses https://github.com/lensfun/lensfun/issues/502

I'm just a bit concerned about naming.


seebk:
From my side the naming is fine and keeps the already lengthy function name short. Would you mind adding the same for TCA and Vignetting? Then the fix would be complete and we could merge it.

PR #2165, opened 2024-02-07
Add Pentax Q-S1

kmilos:
Also prettify all Q series model names.


Lobalobo:
Just a note of support for the request to add Pentax Q-S1.

PR #2169, opened 2024-02-29
Add clang-tidy CI action

### LINKED ISSUE #2168: Setup clang-tidy and clang-format CI pipelines.
In addition to #2164 it would be nice to have clang-tidy and clang-format CI pipelines. (and some additional warning flags can be turned on by default, they are not included into Wall\Wextra)
Unfortunately both of them require significant codebase changes:

- Codestyle is inconsistent across the project and even sometimes strange, so even if it is possible to fine-tune clang-format, it will rewrite a lot of code. But I guess it contribute to code quality and readability and may help to newcomers.
- Only few clang-tidy checks can pass right now, but to make it useful, it may require fixing or rewriting lots of code. But those checks make code more robust and reliable.

CAHEK7:
Partially fixes #2168

It runs actual check only if there are .cpp\.h file changed and does the check only for the changed parts of the files.

By default it enables lots of checks with some exclusions
```
         misc-*,
         -misc-no-recursion,
         -misc-non-private-member-variables-in-classes,
         modernize-*,
         -modernize-use-trailing-return-type,
         -modernize-use-auto,
         readability-*,
         -readability-implicit-bool-conversion,
         -readability-braces-around-statements,
         -readability-identifier-length,
         -readability-magic-numbers,
         -readability-function-cognitive-complexity,
         -readability-uppercase-literal-suffix,
         -readability-simplify-boolean-expr,
         -readability-else-after-return,
         -readability-named-parameter,
         cppcoreguidelines-*,
         -cppcoreguidelines-avoid-magic-numbers,
         -cppcoreguidelines-avoid-goto,
         -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
         -cppcoreguidelines-pro-bounds-pointer-arithmetic,
         -cppcoreguidelines-pro-bounds-constant-array-index,
         -cppcoreguidelines-pro-type-vararg,
         -cppcoreguidelines-pro-type-union-access,
         -cppcoreguidelines-narrowing-conversions,
         -cppcoreguidelines-init-variables,
         -cppcoreguidelines-non-private-member-variables-in-classes,
```
and after the check+exclusions the exception come.
Exclusions should not be fixed I guess, exceptions must be. But it's just my point of view and the list and the order is a subject for discussions.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -1,0 +1,45 @@
+name: Clang-Tidy
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  BUILD_TYPE: Release
+
+jobs:
+  check-needed:
+    runs-on: ubuntu-latest
+
+    outputs:
+      status: ${{ steps.needed.outcome }}
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Checkout base ref
+      run: git fetch origin "${GITHUB_BASE_REF}:refs/remotes/origin/${GITHUB_BASE_REF}";
+    - name: Check if clang-tidy required
+      id: needed
+      continue-on-error: true
+      run: git diff --name-only -U0 "origin/${GITHUB_BASE_REF}" | grep -c -e "\.cpp$" -e "\.h\.in$" -e "\.h\.in\.cmake$" -e "\.h$" -e "\.hpp"
+    
+  run-tidy:
+    needs: check-needed
+    if: needs.check-needed.outputs.status == 'success'
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Checkout base ref
+      run: git fetch origin "${GITHUB_BASE_REF}:refs/remotes/origin/${GITHUB_BASE_REF}";
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_TESTS=ON -DBUILD_LENSTOOL=ON
+    - name: Install clang-tidy for clang-tidy-diff
+      run: sudo apt update && sudo apt install clang-tidy
+    - name: Run clang-tidy
+      run: git diff origin/${GITHUB_BASE_REF}..HEAD | clang-tidy-diff -use-color -p1 -regex .*cpp -path ${{github.workspace}}/build > out 2> err
+    - name: Check results
+      run: if grep -q "warnings treated as errors" err; then cat out; false; else true; fi
--- a/libs/lensfun/.clang-tidy
+++ b/libs/lensfun/.clang-tidy
@@ -1,0 +1,74 @@
+Checks: '-*,
+         clang-analyzer-*,
+         performance-*,
+
+         misc-*,
+         -misc-no-recursion,
+         -misc-non-private-member-variables-in-classes,
+
+         modernize-*,
+         -modernize-use-trailing-return-type,
+         -modernize-use-auto,
+
+         readability-*,
+         -readability-implicit-bool-conversion,
+         -readability-braces-around-statements,
+         -readability-identifier-length,
+         -readability-magic-numbers,
+         -readability-function-cognitive-complexity,
+         -readability-uppercase-literal-suffix,
+         -readability-simplify-boolean-expr,
+         -readability-else-after-return,
+         -readability-named-parameter,
+
+         cppcoreguidelines-*,
+         -cppcoreguidelines-avoid-magic-numbers,
+         -cppcoreguidelines-avoid-goto,
+         -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
+         -cppcoreguidelines-pro-bounds-pointer-arithmetic,
+         -cppcoreguidelines-pro-bounds-constant-array-index,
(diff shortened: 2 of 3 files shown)


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,61 @@
...
+          echo "base=${{ github.event.before }}" >> "$GITHUB_OUTPUT"
+        fi
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_TESTS=ON -DBUILD_LENSTOOL=ON
+    - name: Install clang-tidy for clang-tidy-diff
+      run: sudo apt update && sudo apt install clang-tidy
+    - name: Run clang-tidy
+      run: git diff -U0 ${{ steps.base-ref.outputs.base }} | clang-tidy-diff -use-color -p1 -regex .*cpp -path ${{github.workspace}}/build > out 2> err
+    - name: Check results
+      run: if grep -q "warnings treated as errors" err; then cat out; false; else true; fi
```
jonas-ott, line 61:
```suggestion
      run: if grep -q "warnings\? treated as errors" err; then cat out; false; else true; fi
```
In case there is only one warning

jonas-ott (changes requested):
I'm in favor of adding clang-tidy checks and running them on PRs.

I didn't check which specific checks are enabled/disabled. Before we start a long discussion I would suggest to just enable clang-tidy and adjust later, if necessary.

Could you add a short documentation on how to use clang-tidy locally? CMake can run clang-tidy so maybe we could use that. But with CMake it's probably impossible to just check one commit.

CAHEK7:
Oh, thanks, I forgot about local runs. Will add it.
As for the rules - I've made a set of exclusions and exceptions reflecting the current state of the project, so if you apply clang-tidy on the current master it will pass and won't spam with thousands of errors.

CAHEK7:
Instructions added to the README.md
I guess the easiest way is just to enable it in cmake and perform the checks for all the files during the build.
Also I provided a way to run it manually, or even the checks can be done in IDE, for example VSCode automatically uses appropriate `.clang-tidy` files and highlight all the problems.

jonas-ott:
> As for the rules - I've made a set of exclusions and exceptions reflecting the current state of the project, so if you apply clang-tidy on the current master it will pass and won't spam with thousands of errors.

I get a lot of `misc-const-correctness` warnings so that check should probably be disabled.

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,61 @@
...
+          echo "base=${{ github.event.before }}" >> "$GITHUB_OUTPUT"
+        fi
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_TESTS=ON -DBUILD_LENSTOOL=ON
+    - name: Install clang-tidy for clang-tidy-diff
+      run: sudo apt update && sudo apt install clang-tidy
+    - name: Run clang-tidy
+      run: git diff -U0 ${{ steps.base-ref.outputs.base }} | clang-tidy-diff -use-color -p1 -regex .*cpp -path ${{github.workspace}}/build > out 2> err
+    - name: Check results
+      run: if grep -q "warnings\? treated as errors" err; then cat out; false; else true; fi
```
jonas-ott, line 61: Second round:
```suggestion
      run: if grep -q "warnings\? treated as error" err; then cat out; false; else true; fi
```
Just noticed that I didn't copy the last `s` while testing the command.

README.md
```diff
@@ -104,6 +104,9 @@ The following CMake options can be set (defaults are upper case):
     -DBUILD_FOR_SSE2=off|ON     build with SSE2 optimisaiton
     -DBUILD_DOC=OFF|on          build documentation
 
+    -DCMAKE_CXX_CLANG_TIDY=clang-tidy       run clang-tidy automatically, alongside with a regular compiler
```
jonas-ott, line 107: Because of `WarningsAsErrors: '*'` the build fails with this. I don't see an easy workaround, so I guess we have to live with it. Checking only specific files seems more useful anyway.

jonas-ott (commented):
This should have been included in the previous review.

CAHEK7:
> > As for the rules - I've made a set of exclusions and exceptions reflecting the current state of the project, so if you apply clang-tidy on the current master it will pass and won't spam with thousands of errors.
> 
> I get a lot of `misc-const-correctness` warnings so that check should probably be disabled.

Unfortunately it depends on clang-tidy version. It has been introduced in clang-tidy-15, but my local installation is 14.
Lets stick with clang-tidy-15 for some time keeping everything stable and without any surprises.

README.md
CAHEK7: I'll add explicit clang-tidy version and it won't break anything with our current configs. (with -misc-const-correctness)

CAHEK7:
With clang-tidy-15 and current `.clang-tidy` configs it won't break the build by default.

Next step after this PR is removing or tweaking exceptions from the configs one by one run full build and fix the code.

jonas-ott (approved):
I did a test run with `cppcoreguidelines-avoid-goto` enabled. Only the goto I added in the same PR is reported as error: 
https://github.com/jonas-ott/lensfun/actions/runs/8133353684/job/22224841044 

With `clang-tidy-15` the build runs without errors:
https://github.com/jonas-ott/lensfun/actions/runs/8133401182/job/22224933425

In my opinion this PR can be merged.

PR #2172, opened 2024-03-06
Add Viltrox 20mm F1.8

CAHEK7:
This is full-manual lens and it does not provide any exif information. I've made a name based on the others Viltrox lens, but it may be like a `Viltrox PFU RBMH 20mm f/1.8 ASPH`.


CAHEK7:
@tuxfanx , done

PR #2173, opened 2024-03-06
Add TTArtisan 50mm F1.4 Tilt

CAHEK7:
This is full-manual lens and it does not provide any exif information.


tuxfanx:
...was able to test the profile on some pictures from the internet.

Issue #2161, opened 2024-02-03
Get rid of legacy C code in C++


CAHEK7:
In https://github.com/lensfun/lensfun/issues/2142 I mentioned that the current codebase while being compiled with C++14 contains lots of pure C leftovers.
One of the example uncovered in the following comment https://github.com/lensfun/lensfun/pull/2138#discussion_r1477043594 due to unnecessary C-style reinterpret cast at https://github.com/lensfun/lensfun/blob/20cbe397dc8e11eea8ea0a7ac4ca42b2b2413f34/libs/lensfun/lens.cpp#L1317

- [ ] get rid of C-style casts, replace it with appropriate C++ casts or even remove them completely.
- [ ] consider to use smart pointers wherever possible
- [ ] implement proper move constructors and operators
- [ ] (questionable) use std library instead of glib or own implementations

PR #2176, opened 2024-03-10
replace custom helper templates with C++ STL

jonas-ott:
With C++17 we can replace some custom templates with C++ STL.

In `libs/lensfun/mod-color.cpp` using `std::clamp` is a slight change in behavior. The custom `clampd` did only clamp to `std::numeric_limits<T>::max()` if T is ether `lf_u16` or `lf_u32`, now it's done for every type. To me this seems more like the desired behavior but If someone isn't comfortable with this I can revert the change.

The other changes shouldn't change behavior.


CAHEK7:
Hi @jonas-ott , could we coordinate the changes? I'd like to make some improvements like https://github.com/lensfun/lensfun/pull/2163  but for the other lfTypes.

libs/lensfun/lens.cpp
```diff
@@ -1115,7 +1115,7 @@ static float __vignetting_dist (
     float d1 = 0.1 / distance;
     float d2 = 0.1 / x.Distance;
 
-    return sqrt (square (f2 - f1) + square (a2 - a1) + square (d2 - d1));
+    return sqrt (pow(f2 - f1, 2) + pow (a2 - a1, 2) + pow (d2 - d1, 2));
```
CAHEK7, line 1118: Potentially that may affect performance. `std::pow` is a common algorithm with lots of additional checks, unless it is not optimized by the compiler. And compilers do strange things - gcc and clang are good, but MSVC does not optimize it at all and always call `pow` or `powf` function. Moreover MSVS and clang distinguish `std::pow(x, 2)` and `std::pow(x, 2.f)` and in the first case it casts everything to `double` which is even slower.

There is an `std::hypot` function, which is more precise:
```suggestion
    return hypot(f2 - f1, a2 - a1, d2 - d1);
```
So if we do care about precision, than is should be replaced with `hypot`, but if we care about performance, the old `x*x` can be better than `std::pow(x, 2)`.

jonas-ott: Doesn't look performance critical. Running the tests with coverage the function got 481 hits.

libs/lensfun/lens.cpp
```diff
@@ -11,7 +11,7 @@
 #include <regex>
 #include <string.h>
 #include <locale.h>
-#include <math.h>
+#include <cmath>
```
CAHEK7, line 14: Since this library extensively omits `std::` namespace, this change leads to unnecessary float to double conversion for all the mathematical functions. When I did it for `lfModifier` class, it made coordinates transformation ~10% slower.

Here is an example: https://godbolt.org/z/Ke1M53K96

But it's a matter for standard library and compiler. MSVS produces identical code, but clang and gcc treat mathematical functions without `std::` prefix from `cmath` header as double.

I'm sorry that such a little changes led to the bigger ones.

jonas-ott: Thanks for explaining the effect of the header change, didn't expect the difference in performance.
> I'm sorry that such a little changes led to the bigger ones.

Well, that's how it is.

CAHEK7 (approved):

CAHEK7:
Intentionally or not, but it is related to #2161

PR #2179, opened 2024-03-11
use std::aligned_alloc, remove duplicate macros

jonas-ott:
Since C++17 there is a [std::aligned_alloc](https://en.cppreference.com/w/cpp/memory/c/aligned_alloc) unfortunately it's not supported by Microsoft (see `Notes` on cppreference). Still better than the existing solution.
Also the include logic for `malloc.h` is duplicated (in `common_code.hpp` and in the cpp files) and not neccecary (`cstdlib` should be used).


CAHEK7 (approved):
Looks good.

seebk:
Note that this broke OSX tests because std::aligned_alloc has the strange condition that it may require the size to be a multiple of the alignment. Fixed in #2637.

PR #2213, opened 2024-04-15
Add YI M1

choryuidentify:
Add 'Yi Technology M1', which is Micro 4/3 form-factor Mirrorless Camera.


choryuidentify:
> FWIW, it doesn't seem to me that there is further commitment to the MILC market, this is perhaps ok to add to `misc.xml` instead?

Sure, I replace it into `misc.xml`

PR #2233, opened 2024-06-13
Add distortion correction for Canon RF-S 18-45mm F4.5-6.3 IS STM

dmitryash:
Verified it in Darktable.


tuxfanx:
Distortion data for the lens are already included in the database. Is there a good reason to replace them?

dmitryash:
@tuxfanx Probably there is no reason anymore.

PR #2256, opened 2024-07-30
stop direct execution of setup.py

### LINKED ISSUE #2237: Direct execution of `setup.py` is deprecated, and current syntax fails with modern Python ecosystem
See https://bugzilla.redhat.com/show_bug.cgi?id=2255539 and https://bbs.archlinux.org/viewtopic.php?id=287178 for some background on this.

lensfun's `apps/CMakeLists.txt` calls `python3 setup.py build` and `python3 setup.py install` to try and build and install the Python library. This usage is deprecated and has been since 2021: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary . Doing it with any modern setuptools at minimum produces a deprecation warning:
```
/usr/lib/python3.12/site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
!!
        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.
        Follow the current Python packaging guidelines when building
        Python RPM packages.
        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
        and https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/ for details.
        ********************************************************************************
!!
  self.initialize_options()
/usr/lib/python3.12/site-packages/setuptools/_distutils/cmd.py:66: EasyInstallDeprecationWarning: easy_install command is deprecated.
!!
        ********************************************************************************
        Please avoid running ``setup.py`` and ``easy_install``.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.
        See https://github.com/pypa/setuptools/issues/917 for details.
        ********************************************************************************
!!
```
However, it gets worse, for downstream packagers at least (this may or may not be an issue when building/installing in a "normal" environment, I haven't checked; it would depend whether whatever path cmake decides to install to is on `PYTHONPATH`). With current setuptools (I believe setuptools is the determining factor, anyway) there's a check that the path being installed to is on `PYTHONPATH`, and if it isn't, setuptools throws a snit, doesn't install there, and builds an egg instead:
```
Checking .pth file support in /builddir/build/BUILDROOT/lensfun-0.3.3-8.fc40.x86_64/usr/lib/python3.12/site-packages/
/usr/bin/python3 -E -c pass
TEST FAILED: /builddir/build/BUILDROOT/lensfun-0.3.3-8.fc40.x86_64/usr/lib/python3.12/site-packages/ does NOT support .pth files
bad install directory or PYTHONPATH
You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:
    /builddir/build/BUILDROOT/lensfun-0.3.3-8.fc40.x86_64/usr/lib/python3.12/site-packages/
and your PYTHONPATH environment variable currently contains:
    ''
Here are some of your options for correcting the problem:
* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files
* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)
* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:
  https://setuptools.pypa.io/en/latest/deprecated/easy_install.html#custom-installation-locations
Please make the appropriate changes for your system and try again.
running bdist_egg
```
...which really isn't what we want, and doesn't work. This caused both Fedora and Arch to ship broken lensfun python packages for a while (which is partly our fault, of course, Fedora's packager just went ahead and packaged the egg file without checking whether it actually worked, which was wrong).

The way you're *really* supposed to do Python packaging now is using a `pyproject.toml` file. There's a lot of info on the https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html blog I linked, the key PEPs are https://peps.python.org/pep-0517/ , https://peps.python.org/pep-0518/ , and https://peps.python.org/pep-0621/ (I don't think I missed any), and you can find guides like https://packaging.python.org/en/latest/guides/section-build-and-publish/ . Of course, integrating this into a cmake-based workflow is more complex. I don't know if there's a standard/best practice way to do that.

For downstream, in Fedora I've gone with something similar to what Arch did - we just go into the `app/` directory and run our standard Python package build/install commands from there, which - given there's no `pyproject.toml` - use the info from `setup.py` to build and install a wheel, without invoking it directly. (I also patched out the `setup.py` invocations to avoid the deprecation warnings and the useless egg). I haven't checked what other distros are doing. But obviously this is not upstreamable.

jonas-ott:
Fixes #2237 (Thanks for the detailed bug report)
I've split this into two commits.
The first one does the bare minimum to fix  the deprecation.  Instead of calling `setup.py` directly, the commands described [here](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary) are used.
The second commit gets rid of `setup.py` and uses a `pyproject.toml` which is supposedly the correct way.

Because of a lack of experience with Python build systems, I'm not super confident that everything is done the correct way, and I would be happy if somebody else would take a look at it.


AdamWill:
On an eyeball review this looks good to me, thanks for the fix.

PR #2314, opened 2024-09-14
Add new camera Leica M11-D

Andries-B:
On 12-9-2024 the Leica M11-D was released


Andries-B:
No, I haven't, couldn't find one on internet either. This just seemed logical (looking at the other models), but maybe it is better to have an actual raw file for verification.

PR #2349, opened 2024-10-19
Add vignetting data for Lumix S 100mm F2.8

charles-dyfis-net:
Particularly  at macro distances, vignetting on this lens is quite significant, and in collecting this dataset it seemed at times to change in odd ways. (I do wonder if the data collected is enough to model it properly).

Note that infinity values were actually taken at maximum focus distance (past infinity). Let me know if this invalidates results.


charles-dyfis-net:
The performance I'm getting with the vignetting model installed is not consistently great.

Working theory is that software is not successfully reading focus distance from metadata, and consequently unable to determine which parameters to use. (This seems fairly likely -- as I understand it, Panasonic's raw files only store focus distance in "steps" as part of a field in a proprietary format).

sarunasb:
Please either provide data (and few images for verifying it) you are satisfied with, or upload imgaes to calibration service 

https://wilson.bronger.org/calibration

PR #2369, opened 2024-11-20
Adjust Sigma 10-18mm f/2.8 so that it matches on Fujifilm X, too

horazont:
The Fujifilm version reports with Sigma prefixed, so we add that for matching.


horazont:
Thanks!

PR #2476, opened 2025-05-16
Add filter to TarFile.extract()

sarunasb:
Add `filter='data'` to TarFile extract for security and as required since Python 3.14.


designwithusman9-stack:
video  upsclar toool  please

PR #2513, opened 2025-08-08
Switch from ownCloud to Nextcloud for the calibration images server

bronger:
Hoping for a more modern and robust foundation, this PR switches to Nextcloud for sharing calibration images with the people doing the calibrations.

If this is merged, I will make minor bug fixes directly to master. It should not be a prolem because it only affects a well-isolated part of the Lensfun repository.


bronger:
Withdrawn.

PR #2518, opened 2025-08-11
Add model in E ... format for Tamron A058

TiemenSch:
Metadata from my Sony A7iii for the Tamron 35-150 (A058) was picked up according to a format that exists for other Tamron lenses in the XML, but not yet for the A058. This PR fixes that detection (by Darktable, that is).

Edit: fixed branch name error..

### DIFF
--- a/data/db/mil-tamron.xml
+++ b/data/db/mil-tamron.xml
@@ -730,6 +730,7 @@
     <lens>
         <maker>Tamron</maker>
         <model>TAMRON 35-150mm F/2-2.8 Di III VXD A058Z</model>
+        <model>E 35-150mm F2.0-F2.8 A058</model>
         <model lang="en">Tamron 35-150mm F/2-2.8 Di III VXD</model>
         <mount>Nikon Z</mount>
         <mount>Sony E</mount>
--- a/data/db/mil-tamron.xml
+++ b/data/db/mil-tamron.xml
@@ -729,8 +729,7 @@
 
     <lens>
         <maker>Tamron</maker>
-        <model>TAMRON 35-150mm F/2-2.8 Di III VXD A058Z</model>
-        <model>E 35-150mm F2.0-F2.8 A058</model>
+        <model>Tamron E 35-150mm f/2-2.8 2.0-2.8 Di III VXD A058Z</model>
         <model lang="en">Tamron 35-150mm F/2-2.8 Di III VXD</model>
         <mount>Nikon Z</mount>
         <mount>Sony E</mount>


TiemenSch:
I'm not an expert, but I thought that might be how its supported on Nikon Z mount. I don't want to break detection for anyone else. I'll gladly remove it if you're sure!

sarunasb:
@TiemenSch I don't think more than one `<model>` element is supported (despite DTD). You can test your local repository with `lenstool -L…`

TiemenSch:
Hmm. It seems as if there have also been commits the other way around: https://github.com/lensfun/lensfun/blob/01d8a200dc2679fc8c236e76467c6215db14218a/data/db/mil-tamron.xml#L519

TiemenSch:
Scrolling through the Tamron file it seems as if Sony and Nikon report the lens names slightly different. Is there a middle ground that gets them both identified well enough? Or should their entries actually be split into two dedicated entries for that?

kmilos:
> Scrolling through the Tamron file it seems as if Sony and Nikon report the lens names slightly different.

This is not restricted to Tamron. There is at least the case for Viltrox as well (see misc.xml for different Z/E/X mount strings), and probably others like Sigma...

> Is there a middle ground that gets them both identified well enough?

Try and see. Perhaps `35-150mm F2-2.8 A058` works?

Lensfun really needs aliasing support...

TiemenSch:
I can get it to work for me using the string that I added. 
However, I'm afraid that removing the other line will break detection for Nikon Z users (which I can't test as I don't have a RAW from a Nikon camera with the equivalent lens).

Edit: just realized what you meant with the `lenstool -L ...` test.

It really is quite unforgiving with the `E` in front for Sony and the differences between general formatting of the F numbers and type Tamron lens type indicators.

After some experimentation, I can only conclude that there is no middle ground.

sarunasb:
`lenstool -L` would help in checking how Lensfun “digests” and matches strings from Exif in RAWs with what it has in its XML DB.

Here is from https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/calibration.rst

>Often, you may want to fine-tune the lens model name. Lensfun normalises names before any matching, so you have >some freedom. For example, upper/lowercase can be changed arbitrarily. Any single “f” is ignored, so you may change >10-18mm 2.8 into 10-18mm f/2.8. If there was a tele converter involved, you must add “converter” into the name so that >Lensfun does not try to derive allowed focal lengths from the lens name. Ordering of parts in the lens name is completely >unimportant for matching. As are single punctuation characters. You may even add things (e.g. 10-18 into 10-18mm) but >be conservative here. Never drop something from what exiv2 says (except for punctuation); this would thwart matching.

It would help to have RAWs from at least Canon, Nikon, Sony for the same Tamron lens. Then experiment with `lenstool`.

sarunasb:
The issue (besides missing alias support) here is that not only E is added, but also decimal notation is used for f/2. So having something like
```
Tamron E 35-150mm f/2-2.8 2.0-2.8 Di III VXD A058Z
```
might work for multiple camera mounts/RAWs. At least seems to work (correct lens is found) for me while querying with:

```
E 35-150mm 2.0-2.8 A058
35-150mm 2-2.8 A058
Tamron 35-150mm F2-2.8
```

tuxfanx:
@sarunasb 
> Tamron E 35-150mm f/2-2.8 2.0-2.8 Di III VXD A058Z

Seems to work with both available lens mounts (Nikon Z, Sony E). However, I was only able to test it for the Nikon Z with a few JPEGs from the internet.

@TiemenSch
On my Sony raw image sample (150mm, f/2.8), the Lensfun vignetting correction seems to be much too strong. Can you confirm this?

TiemenSch:
> The issue (besides missing alias support) here is that not only E is added, but also decimal notation is used for f/2. So having something like
> 
> ```
> Tamron E 35-150mm f/2-2.8 2.0-2.8 Di III VXD A058Z
> ```
> 
> might work for multiple camera mounts/RAWs. At least seems to work (correct lens is found) for me while querying with:
> 
> ```
> E 35-150mm 2.0-2.8 A058
> 35-150mm 2-2.8 A058
> Tamron 35-150mm F2-2.8
> ```

That works! Makes sense when I look at the rules in hindsight, though it feels somewhat unintuitive that specifying more of the variations results in a broader match. Will update PR!

TiemenSch:
> @sarunasb
> 
> > Tamron E 35-150mm f/2-2.8 2.0-2.8 Di III VXD A058Z
> 
> Seems to work with both available lens mounts (Nikon Z, Sony E). However, I was only able to test it for the Nikon Z with a few JPEGs from the internet.
> 
> @TiemenSch On my Sony raw image sample (150mm, f/2.8), the Lensfun vignetting correction seems to be much too strong. Can you confirm this?

I think you're right! Not sure if this is due to differences between E/Z mount versions, but I agree that it's too strong. That should probably be a separate issue.

tuxfanx:
> > @sarunasb
> > > Tamron E 35-150mm f/2-2.8 2.0-2.8 Di III VXD A058Z
> > 
> > 
> > Seems to work with both available lens mounts (Nikon Z, Sony E). However, I was only able to test it for the Nikon Z with a few JPEGs from the internet.
> > @TiemenSch On my Sony raw image sample (150mm, f/2.8), the Lensfun vignetting correction seems to be much too strong. Can you confirm this?
> 
> I think you're right! Not sure if this is due to differences between E/Z mount versions, but I agree that it's too strong. That should probably be a separate issue.

Thanks, you are right, this should be a new issue.
Note:
In addition to the classic lensfun profiling method, there is also a [Python script](https://github.com/uchrisu/lf_fitexif) for Sony E lenses that converts the manufacturer's correction data into a lensfun profile. The data obtained this way may yield better results.

PR #2524, opened 2025-08-15
Prettyfy display names leica

Andries-B:
As discussed/proposed in https://github.com/lensfun/lensfun/discussions/2523
This PR does
```
    NOT TOUCH the <model>string</model> line
    prettify some <model lang="en">string</model> lines

```
E.g.

- `<model>APO-Summicron-SL 1:2/35 ASPH.</model>` is literal as reported by exiv2, and is how Leica refers to the lens. However, look for instance at the Leica Store website (Amsterdam, https://www.leicastoreamsterdam.nl/copy-of-leica-apo-summicron-sl-35-f-2-asph-black.html) where they use a different format
- `<model lang="en">APO-Summicron-SL 35mm f/2 ASPH.</model>` is prettified

### DIFF
--- a/data/db/mil-canon.xml
+++ b/data/db/mil-canon.xml
@@ -193,6 +193,14 @@
 
     <camera>
         <maker>Canon</maker>
+        <model>Canon EOS R5 C</model>
+        <model lang="en">EOS R5 C</model>
+        <mount>Canon RF</mount>
+        <cropfactor>1</cropfactor>
+    </camera>
+
+    <camera>
+        <maker>Canon</maker>
         <model>Canon EOS R5m2</model>
         <model lang="en">EOS R5 Mark II</model>
         <mount>Canon RF</mount>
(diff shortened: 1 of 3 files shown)


tuxfanx:
...and others would like to see the lens designations in the notation used by the manufacturers on its lenses or on its official websites (e.g. see Leica-Camera.com). The notation on the websites depends on the language of the site. The notation on the lenses not. 
[Cropfactor for M8 is already 1.33](https://github.com/lensfun/lensfun/blob/e289233ba6774682358534f4d3bdd78cd6275a06/data/db/rf-leica.xml#L31)

Andries-B:
As per discussion https://github.com/lensfun/lensfun/discussions/2523 :
The PR for Leica lenses goes against this best practice. I agree with you, that it is best to follow the manufacturers notation (unless ...etc).

Please cancel (do not accept) this PR

Andries-B:
As per discussion https://github.com/lensfun/lensfun/discussions/2523 :
The PR for Leica lenses goes against this best practice. I agree with you, that it is best to follow the manufacturers notation (unless ...etc).

Please cancel (do not accept) this PR

PR #2525, opened 2025-08-15
Prettify display names sigma

Andries-B:
As discussed per https://github.com/lensfun/lensfun/discussions/2523


tuxfanx:
f/XX does not correspond to the[ manufacturer's notation](https://www.sigma-global.com/en/lenses/). Sigma mostly uses FXX.

Andries-B:
As per discussion https://github.com/lensfun/lensfun/discussions/2523 :
This PR for Sigma lenses is withdrawn/cancelled, until it has become clear what is needed.

Please cancel (do not accept) this PR

PR #2526, opened 2025-08-15
Prettyfy display names panasonic

Andries-B:
As discussed per https://github.com/lensfun/lensfun/discussions/2523

PR #2538, opened 2025-08-21
Prettify mil-nikon.xml

kmilos:
By the by, I still don't feel what was done in slr-nikon.xml was the best choice, there was nothing wrong w/ [Nikon product names "AF-S Nikkor" and the like](https://imaging.nikon.com/imaging/lineup/lens/f-mount/)... Just removing "Nikon" in front would've been my preference.

Cf. also https://www.dpreview.com/products/nikon/lenses?subcategoryId=lenses&page=1

### DIFF
--- a/data/db/mil-nikon.xml
+++ b/data/db/mil-nikon.xml
@@ -385,7 +385,7 @@
 
     <lens>
         <maker>Nikon</maker>
-        <model>NIKKOR Z MC 50mm f/2.8</model>
+        <model>Nikkor Z MC 50mm f/2.8</model>
         <mount>Nikon Z</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>
@@ -407,7 +407,7 @@
 
     <lens>
         <maker>Nikon</maker>
-        <model>NIKKOR Z 14-24mm f/2.8 S</model>
+        <model>Nikkor Z 14-24mm f/2.8 S</model>
         <mount>Nikon Z</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>
@@ -477,7 +477,7 @@
 
     <lens>
         <maker>Nikon</maker>
-        <model>NIKKOR Z 14-30mm f/4 S</model>
+        <model>Nikkor Z 14-30mm f/4 S</model>
         <mount>Nikon Z</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>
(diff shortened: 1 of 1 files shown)


tuxfanx:
[I removed "Nikon" from slr-nikon.xml!?](https://github.com/lensfun/lensfun/commit/fb37642968522a8c49823a6ce2ee8bac37c44e37) And "NIKKOR" was changed to "Nikkor" according to the prevailing model name notation. Why didn't you object to the prevailing notation much earlier?

For the Nikkor Z lens series, I intentionally left "NIKKOR" as it is because that's how the model names appear in the EXIV data, and that was already the prevailing notation. [In my opinion, "NIKKOR" could therefore remain](https://github.com/lensfun/lensfun/pull/2526#issuecomment-3200353497) as it is for the Z lens series. But, as we all know, taste is subjective.

kmilos:
> Why didn't you object to the prevailing notation much earlier?

[I did, right after the first commit](https://github.com/lensfun/lensfun/commit/fb37642968522a8c49823a6ce2ee8bac37c44e37#comments).

(One downside to direct commits is that you rarely get feedback as there is no chance for others to review a proper PR.)

> I intentionally left "NIKKOR" as it is because that's how the model names appear in the EXIV data

So is LUMIX, yet you went w/ Lumix everywhere. (TAMRON is the same btw.)

As there is no real plan here and everything seems to boil down to personal preference in a given moment, there's no point to proceeding it would seem.

sarunasb:
We should use ‘Nikkor’, IMO. They use all uppercase in some occasions for visual emphasis (on lenses, where print is already small, for example), but grammatically, in the database of names, it is not quite right.

tuxfanx:
@kmilos 
Sorry, I missed your earlier comment.
I can agree with @sarunasb 
So @kmilos is welcome to reopen his PR. I will always try to adapt new profiles to the prevailing notation. If we try to follow the manufacturers' notation, where there are sometimes differences between their datasheets and different websites, it will remain a perpetual issue.

sarunasb (approved):

kmilos:
> If we try to follow the manufacturers' notation, where there are sometimes differences between their datasheets and different websites, it will remain a perpetual issue.

True, especially if there is nothing in Exif to guide us and one relies solely on web sites. However, one can clearly see some (if not perfect) consistency w/ Nikon F lenses:

"AF-S|AF-P [DX] [Zoom-]Nikkor" etc.

Similarly, Lensfun can pick _one_ style per manufacturer/line that _mostly_ aligns w/ product marketing to avoid confusion but still looks pretty for the database presentation.

tuxfanx (approved):

PR #2539, opened 2025-08-21
Fix Sigma 16-300mm F3.5-6.7 DC OS | C

kmilos:


tuxfanx:
@kmilos thanks,
initialy I was confused because it was stored as `Canon 16-300mm F3.5-6.7 DC OS | Contemporary 025` on the server and exiftool reported the ID in the same way. That leads to this mistake.
Canon 16-300mm F3 5-6 7 DC OS | Contemporary 025--016mm--8 0_16mm

kmilos:
> exiftool reported the ID in the same way

Ah, you should rarely rely on the "Composite" result, it is just [an extrapolation/summary exiftool makes, not a real tag](https://exiftool.org/TagNames/Composite.html) present in the file:
```
[ExifIFD]       LensInfo                        : 16-300mm f/0
[ExifIFD]       LensModel                       : 16-300mm F3.5-6.7 DC OS | Contemporary 025
[Canon]         LensModel                       : 16-300mm F3.5-6.7 DC OS | Contemporary 025
[Canon]         LensType                        : Unknown (4976)
[Composite]     LensID                          : Canon 16-300mm F3.5-6.7 DC OS | Contemporary 025
```
I guess in this case it fell back to using the body Make as LensMake was not present and the lens is not in the exiftool internal translation table...

PR #2540, opened 2025-08-22
Model names Leica lenses

Andries-B:
About model names of Leica lenses. `Exiv2 `reports these tags: `APO-SUMMICRON-SL 1:2/35 ASPH.` `APO-SUMMICRON 1:2/43 ASPH.`
Apparently the exif-tags are reported in all-caps.

Counting the names of Leica lenses in the lensfun-database result: 
9 names: like APO-Summicron-SL 1:2/35 ASPH.
3 names: like Elmarit-M 90mm f/2.8

The notation in the lensfun-database of the 9 lenses is almost equal to the exif-tag, except for character-case. The 3 other lenses use an f/xx style.

Here is my proposal:
Since lens matching is case-insensitive and after reading https://github.com/lensfun/lensfun/pull/2526 :
* for Leica lenses
* simplify the database by only using the exif-tag in `<model>` and removing `<model lang="en">`
* exception: some models are extended by e.g. (11879/11882). There are many generations of Summicrons that are identified by this number. For this `<model lang="en">` is used
* change case according to
-- Acronym type abbreviations: APO not Apo
-- Title case: Summicron, Summilux, Elmarit, Vario-Elmarit
-- Abbreviation: Asph. not ASPH.
-- Apperture and focallength: fractional notation, e.g. 1:2.8, followed by /focallength

### DIFF
--- a/data/db/compact-leica.xml
+++ b/data/db/compact-leica.xml
@@ -171,8 +171,7 @@
     <lens>
         <maker>LEICA CAMERA AG</maker>
         <maker lang="en">Leica</maker>
-        <model>APO-SUMMICRON 1:2/43 ASPH.</model>
-        <model lang="en">APO-Summicron 43mm f/2 ASPH.</model>
+        <model>APO-Summicron 1:2/43 Asph.</model>
         <mount>leicaQ343</mount>
         <cropfactor>1.0</cropfactor>
         <calibration>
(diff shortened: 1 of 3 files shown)


sarunasb:
Looks good to me.

Andries-B:
> Shouldn't we also omit the "mm" indication for the focal length in the display names for the `Summicron-M 1:2/35mm Asph. `and `APO-Summicron-M 1:2/75mm Asph.` ?

Yes, I missed that, thanks! Now included in this PR

PR #2542, opened 2025-08-23
Correct Panasonic lens names

Andries-B:
* some white space formatting
* swapped tags `<model>` and `<model lang="en">`
* added Asph., Mega, Power (not power) where appropriate
* corrected lens names and checked with panasonic.com

On a side note: I would like to see the lens name EXIF-tag as-is in the `<model>` line; maybe in some cases allow different capitalization. But I don't have that information.

### DIFF
--- a/data/db/mil-panasonic.xml
+++ b/data/db/mil-panasonic.xml
@@ -522,7 +522,7 @@
     <lens>
         <maker>Panasonic</maker>
         <model>Lumix G X Vario PZ 14-42mm f/3.5-5.6 Asph. Power OIS</model>
-        <model lang="en">Lumix G X Vario PZ 14-42mm f/3.5-5.6 Asph. Power OIS</model>
+<!--    <model lang="en">Lumix G X Vario PZ 14-42mm f/3.5-5.6 Asph. Power OIS</model>  -->
         <mount>Micro 4/3 System</mount>
         <focal min="14" max="42"/>
         <aperture min="3.5" max="22"/>
@@ -666,7 +666,7 @@
     <lens>
         <maker>Panasonic</maker>
         <model>Lumix G Vario 14-42mm f/3.5-5.6 II Asph. Mega OIS</model>
-        <model lang="en">Lumix G Vario 14-42mm f/3.5-5.6 II Asph. Mega OIS</model>
+<!--    <model lang="en">Lumix G Vario 14-42mm f/3.5-5.6 II Asph. Mega OIS</model>  -->
         <mount>Micro 4/3 System</mount>
         <focal min="14" max="42"/>
         <aperture min="3.5" max="5.6"/>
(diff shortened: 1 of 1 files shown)


Andries-B:
Nice find! Updated the PR

tuxfanx:
> On a side note: I would like to see the lens name EXIF-tag as-is in the `<model>` line; maybe in some cases allow different capitalization. But I don't have that information.

[Different capitalization should be possible](https://wilson.bronger.org/lensfun/el_lens.html)

sarunasb:
There might have been a reason why e.g. ‘Asph. Power OIS’ were in `<model>` element (which is used for matching). 

If that is what is returned by exiv2, and not present in `<model>`, then matching might fail.

Checking exiv2 source, I don't see lens ID to Lens model translation for Panasonic/Panasonic. Panasonic cameras might put the literal string into Exif data. However Olympus cameras, which use the same µ4/3 Lumix lenses, puts lens ID into Exif. So exiv2 translates those into lens model. And in `src/olympusmn_int.cpp` the translation does:
```
      {{2, 35, 16}, "Lumix G Vario 35-100mm F4.0-5.6 Asph. Mega OIS"},
```
i.e. ‘… Asph. Mega OIS’ is in exiv2 output.

Andries-B:
Than, I made the wrong **assumptions** (always dangerous) about what is in the exiv2 output and didn't realise that maybe different camera brands write different exif information about the same lens. Lens detection should not be broken, so I will cancel the swapping of the `<model>`.

Andries-B:
Now that the PR has been edited (undo the swapping of tags), the PR is a much smaller and safer, with no changes to the `<model>` line.

tuxfanx:
I don't understand why the additional designations "Asph.," "Mega OIS," and "Power OIS" should not be in the display names. Just because the headings on Panasonic's English website don't list them? These additional designations can be found on other websites, including Panasonic's, and on the lenses themselves. The additional designations are also handled differently in the model names in the mil-panasonic.xml file. Wouldn't it be better if we used them consistently in the display names?
It makes the different models easier to distinguish in their names.


Lumix G X Vario PZ 14-42

#############

DE-Lumix G X Vario PZ 14-42

Andries-B:
I think so too. The initial method I used in order to get that result was not the right method, because that touched the `<model>` line. Next I'll change it in the `<model lang="en">` line only.

Andries-B:
Summary of what this PR does now:

* some white space formatting
* do not touch `<model>`
* added Asph., Mega, Power (not power) where appropriate
* corrected lens names and checked with panasonic.com

Andries-B:
If this change is agreed upon, then we (or I) can simplify (= remove the double) the lines where `<model>` and `<model lang="en">` are equal.
E.g.
`<model>Lumix G Vario 45-150mm f/4.0-5.6 Asph. Mega OIS</model>`

`<model lang="en">Lumix G Vario 45-150mm f/4.0-5.6 Asph. Mega OIS</model>`

Andries-B:
Summary of what this PR does now:

* some white space formatting
* do not touch `<model>`
* remove redundant `<model lang="en">` lines
* add Asph., Mega, Power (not power) where appropriate
* correct lens names and checked with panasonic.com

PR #2549, opened 2025-08-25
Fix Canon RF naming style for consistency

kmilos:
The Exif itself (and Canon marketing) uses the "F" rather than "f/" notation for the recent RF series, as do most lenses in the database already.


tuxfanx (commented):
Should it not be `RF 24mm F1.4 L VCM `?

kmilos:
It is.

PR #2556, opened 2025-08-27
Replace 'мм' with regular 'mm'

Andries-B:
Replace 'мм' with regular 'mm'

### DIFF
--- a/data/db/slr-ussr.xml
+++ b/data/db/slr-ussr.xml
@@ -152,7 +152,8 @@
         <maker>KMZ</maker>
         <maker lang="ru">КМЗ</maker>
         <model>MC Helios-44M-4 58mm 1:2</model>
-        <model lang="ru">МС Гелиос-44М-4 58mm 1:2</model>
+        <model lang="en">MC Helios-44M-4 58mm 1:2</model>
+        <model lang="ru">МС Гелиос-44М-4 58мм 1:2</model>
         <mount>M42</mount>
         <cropfactor>1.538</cropfactor>
         <calibration>
@@ -180,7 +181,8 @@
         <maker>KMZ</maker>
         <maker lang="ru">КМЗ</maker>
         <model>MC Helios-44M-4 58mm 1:2</model>
-        <model lang="ru">МС Гелиос-44М-4 58mm 1:2</model>
+        <model lang="en">MC Helios-44M-4 58mm 1:2</model>
+        <model lang="ru">МС Гелиос-44М-4 58мм 1:2</model>
         <mount>M42</mount>
         <cropfactor>1</cropfactor>
         <calibration>
@@ -208,7 +210,8 @@
         <maker>KMZ</maker>
         <maker lang="ru">КМЗ</maker>
         <model>Helios-44 58mm 1:2</model>
-        <model lang="ru">Гелиос-44 58mm 1:2</model>
+        <model lang="en">Helios-44 58mm 1:2</model>
+        <model lang="ru">Гелиос-44 58мм 1:2</model>
         <mount>M42</mount>
         <cropfactor>1</cropfactor>
         <calibration>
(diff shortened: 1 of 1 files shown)


sarunasb:
Those are already lowercase ‘m’ in moscovian. It seems they are translating SI into their own...

bronger:
Unless a native speaker says otherwise, I would leave the “мм”.

Andries-B:
On the list of supported lenses ( https://lensfun.github.io/lenslist/ and https://wilson.bronger.org/lensfun_coverage.html ) appears a line "MTO" "МТО-500 500мм f/8". In my eyes, the lenslist uses international notation, or so one could say lang="en", not lang="ru". That was my starting point for changing ALL occurances of "мм".

Andries-B:
Just a thought: perhaps https://github.com/lensfun/lensfun/pull/2559 fixes the fact that "мм" is appearing on the lenslist. If so, than 'mm' is displayed in english/international pages, 'мм' is on the lang="ru" pages.

Andries-B:
> Unless a native speaker says otherwise, I would leave the “мм”.

I changed the PR: instead of changing the “мм”, it is now left as-is in the `<model lang='ru'>` line; added a `<model lang='en'>` line with 'mm'

bronger:
Possibly I misread the original PR. The current commit is a no-op because the lens name without any language attached is the fallback anyway. (Unless code using the DB is buggy.)

Andries-B:
On the list of supported lenses ( https://lensfun.github.io/lenslist/ and https://wilson.bronger.org/lensfun_coverage.html ):
KMZ - МИР-1B 2.8/37
KMZ - ЮПИТЕР-37AM MC 3.5/135
MTO- МТО-500 500мм f/8
The model-names are English not Russian in the `slr-ussr.xml`, but what seems to appear is the `<model lang='ru'>` description

Perhaps the code is buggy, because another observation, in `slr-ussr.xml`, but not in the list, eg.:
KMZ - MC Variozenitar
KMZ - MC APO Telezenitar
BelOMO - MC Peleng

sarunasb:
Lenslist generation script needs to be checked.

Andries-B:
> Lenslist generation script needs to be checked.

Would that be the file `tools/lenslist/show_lensfun_coverage.py` ? I would like to change and test the `find_best` function before creating another PR. Are there any pointers/instructions on using `tools/lenslist/show_lensfun_coverage.py` ?

sarunasb:
Yes, it's `tools/lenslist/show_lensfun_coverage.py`. It is run by GitHub Actions every time there is a commit. Also pushed to https://github.com/lensfun/lensfun.github.io automatically.

sarunasb:
This in Lens class returns the shorter string.  That's why "ru" entry appears for Mir-1B full-frame option.

```
    def __lt__(self, other):
        return (self.maker, self.model, self.crop) < (other.maker, other.model, self.crop)
 ```
Changing `МИР-1B 2.8/37` to `МИР-1В 37мм f/2.8` "fixes" it.

bronger:
Odd … I thought I made `show_lensfun_coverage.py` but “blame” says otherwise? Anyway, is https://github.com/lensfun/lensfun/blob/master/tools/lenslist/show_lensfun_coverage.py#L20-L22 the problem? In absence of am `"en"` lens name, it takes the first in alphabetical oder.

sarunasb:
> Anyway, is https://github.com/lensfun/lensfun/blob/master/tools/lenslist/show_lensfun_coverage.py#L20-L22 the problem?

Sort of. It would be better is search is done for models, where either `lang="en"`  or `lang` is empty, not something else.

bronger:
I think one can safely assume that there is either a name without any language attribute, or we have to return `None` anyway. So just append `if not element.attrib.get("lang")` to line 20.

sarunasb:
> So just append `if not element.attrib.get("lang")` to line 20.

Yes, that does it.

Andries-B:
Yes, that does it. So, this PR can be discarded since `show_lensfun_coverage.py` will be fixed.

PR #2557, opened 2025-08-27
Replace '×' with regular 'x'

Andries-B:
Replace '×' with regular 'x'

### DIFF
--- a/data/db/compact-canon.xml
+++ b/data/db/compact-canon.xml
@@ -1472,8 +1472,8 @@
     <lens>
         <maker>Canon</maker>
         <model>Canon PowerShot S70 &amp; compatibles, with TC-DC10 2x converter</model>
-        <model lang="en">fixed lens, with TC-DC10 2× converter</model>
-        <model lang="de">festes Objektiv, mit TC-DC10 2×-Konverter</model>
+        <model lang="en">fixed lens, with TC-DC10 2x converter</model>
+        <model lang="de">festes Objektiv, mit TC-DC10 2x-Konverter</model>
         <mount>canonS70</mount>
         <cropfactor>4.843</cropfactor>
         <aspect-ratio>4:3</aspect-ratio>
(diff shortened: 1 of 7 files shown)


tuxfanx:
The special character could certainly be used in the display name. Personally, though, I would prefer the regular "x". 
But the special character seems to have been used more frequently. I don't care what the final choice is.

sarunasb:
It might be that in XML it would be necessary to use an entity notation for × — `&times;` Like it is done already for an ampersand & (`&amp;`), for example. Not sure how this affects applications using Lensfun.

bronger:
I have not looked at it, but it was probably me. It is well-formed XML. I don’t care much, but the “×” are the accurate symbol in this case. Do they cause trouble?

Andries-B:
> I have not looked at it, but it was probably me. It is well-formed XML. I don’t care much, but the “×” are the accurate symbol in this case. Do they cause trouble?

No, they do not cause trouble, not that I'm aware of. This PR was inspired by https://github.com/lensfun/lensfun/pull/2542#pullrequestreview-3147494165 . I don't care much either, maybe just a tiny little bit :-)  in my opinion x is simpler than × because the symbol is on all keyboards.

sarunasb:
Perhaps leave the proper multiplication symbol then `×` then. Compose key + x + x.

Andries-B:
I guess someone has to make the choice:
1. use the proper symbol `×` , which is well-formed XML and which is not standard on all keyboards (workaround for Linux is Compose key + x + x, Compose key is not standard available on Windows, MacOS)
2. use the regular `x`, close enough to `×`, is a key on all keyboards, easy to type

Probably everyone can live with any outcome, and there are only very few occurances of this symbol.

Andries-B:
Here we are:
don't fix it if it isn't broken
do the right thing, use the proper &times; symbol

So, discard and close this PR

PR #2558, opened 2025-08-27
Use dot as decimal separator

Andries-B:
1,4 -> 1.4
2,8 -> 2.8
3,4 -> 3.4
3,5 -> 3.5

### DIFF
--- a/data/db/mil-sigma.xml
+++ b/data/db/mil-sigma.xml
@@ -1038,7 +1038,7 @@
     <lens>
         <maker>Sigma</maker>
         <model>Sigma 28mm F1.4 DG HSM | A</model>
-        <model lang="en">28mm F1,4 DG HSM | Art</model>
+        <model lang="en">28mm F1.4 DG HSM | Art</model>
         <mount>Nikon F AF</mount>
         <mount>Leica L</mount>
         <mount>Sony E</mount>
@@ -1068,7 +1068,7 @@
      <lens>
         <maker>Sigma</maker>
         <model>Sigma 28mm F1.4 DG HSM | A</model>
-        <model lang="en">28mm F1,4 DG HSM | Art</model>
+        <model lang="en">28mm F1.4 DG HSM | Art</model>
         <mount>Nikon F AF</mount>
         <mount>Leica L</mount>
         <mount>Sony E</mount>
--- a/data/db/misc.xml
+++ b/data/db/misc.xml
@@ -333,7 +333,7 @@
 
     <lens>
         <maker>Voigtländer</maker>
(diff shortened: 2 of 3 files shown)


sarunasb:
Somewhat curious, how did comma as decimal separator appeared here. Calibration done on localised system and exiv2 returning localised notation?

tuxfanx:
> Somewhat curious, how did comma as decimal separator appeared here. Calibration done on localised system and exiv2 returning localised notation?

Most of them are purely manual lenses.

bronger:
The advantage of the comma is: It represents the accurate label printed on the lens.

The advantage of the dot is: If somebody makes regex matching, they must cover the case that dot or comma might be the correct decimal separator.

But then, matching lens names is very difficult anyway. Lensfun itself strips all punctuation before it does matches.

Without use cases that have concrete issues with one thing or the other, it is difficult to decide. I personally don’t care.

Andries-B:
In my eyes, the list of supported lenses ( https://lensfun.github.io/lenslist/ and https://wilson.bronger.org/lensfun_coverage.html ) uses international notation, or so one could say lang="en". In US-English and in UK-English the decimal separator is a dot.

There are differences how brands name their lenses on their websites/documentation and what is inscripted on the physical lens and what is in the EXIF-tag as returned by exiv2 (see e.g. discussion https://github.com/lensfun/lensfun/discussions/2523 ) However, one conclusion was that it is best to follow the manufacturers notation (unless ...etc).

For this PR: Sigma uses a dot. The other lenses might exchange data or not (unknown to me).

This PR tries to improve consistency of the notation in the lens database. This PR does not solve any problem, because there is no problem. Lensfun itself is invariant to dot or comma. If localization is needed the comma can be added with a `<model lang="de">` or `<model lang="nl">`

tuxfanx:
As far as I can tell, the Voigtländer and Zeiss lenses with the comma in their model names are purely processor-less, manual lenses. The model name was manually adapted to the engraving on the lens.

I would like to advocate for the consistent use of the decimal point for aperture values ​​for the following reasons:

- In the EXIF ​​data of modern Zeiss and Voigtländer lenses with processors the decimal point is used for decimal separation in the exif data.

- Some profiles of the other older, manual Zeiss Jena lenses already contain a decimal point, despite their engraving on the lenses. If we decide to use the comma, these and perhaps other lenses should have to be adapted.

- When creating profiles (lenses.txt), the use of the dot is required and that is eliminating the need to manually adjust the generated XML profile.

- The use of the dot is more in line with the commonly used notation for apertures.

- Consistent use of the dot simplifies filtering and searching in lists and files.

PR #2561, opened 2025-08-29
Remove lenses without calibration data

Andries-B:
This PR removes 7 lenses from the xml files. These are lenses that have no calibration data (no distortion, no tca, no vignetting data). The lenses are not counted nor shown in the supported lenses list, because there is no calibration data.

However, if there is a reason for keeping these placeholders, please let me know and this PR will be discarded.


Andries-B:
Found another lens without calibration data, so this PR removes 8 lenses.

In generic.xml there are 3 generic lenses, without calibration data. These are untouched.

sarunasb:
I don't know why those lenses are in the database. Git history doesn't seem to go that far, so pre-Git, unless I'm missing something. Just a guess, but maybe at some point Lensfun was envisioned as aslo a historical database of every lens ever made. It's clear it has never became such. There should be no harm in removing those lenses, as far as I can tell. Tey will still remain in Git, if anything.

bronger:
Sorry for being too late. This also removes the cropfactor, which is needed if you change the projection.

For instance: You made a wide-angle photo of people. Those at the edges have distorted faces. In e.g. Darktable, you change projection to “stereographic”, and all faces look good. However, this does not work for the deleted lenses anymore.

sarunasb:
On 2025-08-31 7:38 AM, Torsten Bronger wrote:
> *bronger* left a comment (lensfun/lensfun#2561) <https://github.com/ 
> lensfun/lensfun/pull/2561#issuecomment-3240074947>
> 
> Sorry for being too late. This also removes the cropfactor, which is 
> needed if you change the projection.
> 
> For instance: You made a wide-angle photo of people. Those at the edges 
> have distorted faces. In e.g. Darktable, you change projection to 
> “stereographic”, and all faces look good. However, this does not work 
> for the deleted lenses anymore.

I looks like this is about Pentax-DA 55-300mm f/4-5.8 ED, which has 
entries for KAF, KAF2 and KAF3 mounts, with different crop factors. The 
rest of removed lenses do not have other variants with calibration data.

So this can be restored
```
    <lens>
         <maker>Pentax</maker>
         <model>smc Pentax-DA 55-300mm f/4-5.8 ED</model>
         <mount>Pentax KAF</mount>
         <cropfactor>1.522</cropfactor>
         <calibration>
         </calibration>
     </lens>
```

tuxfanx:
>Sorry for being too late. This also removes the cropfactor, which is
 needed if you change the projection.

I don't understand the point of all this. I just feel like this is an attempt to fix a problem that's not part of lensfun's mission.

bronger:
@sarunasb @tuxfanx Sorry that you received my comment by mail. I had deleted it after seconds. I mixed up `<lens>` crop factor with `<camera>` crop factor. Sorry for the noise!

PR #2562, opened 2025-08-30
Restore times symbol in lens display names instead of regular x

Andries-B:
Restore × (times symbol) in English and German lens display names instead of regular x.

Make display names consistent.
Don't touch `<model>` line.
Edit `<model lang="en">` only.

See also https://github.com/lensfun/lensfun/pull/2557


sarunasb:
Looks good.

PR #2564, opened 2025-08-31
Restore previously deleted lens "smc Pentax-DA 55-300mm f/4-5.8 ED"

Andries-B:
Restore lens smc Pentax-DA 55-300mm f/4-5.8 ED (was deleted in https://github.com/lensfun/lensfun/pull/2561 )

### DIFF
--- a/data/db/slr-pentax.xml
+++ b/data/db/slr-pentax.xml
@@ -1219,6 +1219,15 @@
 
     <lens>
         <maker>Pentax</maker>
+        <model>smc Pentax-DA 55-300mm f/4-5.8 ED</model>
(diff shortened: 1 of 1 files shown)


tuxfanx:
I don't understand the point of all this (#2561). I just feel like this is an attempt to fix a problem that's not part of lensfun's mission.
Is there an explanation for why this is part of lensfun's workload?
If so, there are probably a few other "empty" profiles that need to be added.

Andries-B:
See comments/discussion in https://github.com/lensfun/lensfun/pull/2561 . This PR is not needed, so closing it.

PR #2565, opened 2025-09-01
Fix fixed lenses' listing in lenslist

sarunasb:
In [lenslist](https://lensfun.github.io/lenslist/) there are some fixed lenses, which are only listed anonymously, as ‘fixed lens.’ Lenslist script assumes all the fixed lenses are in `compact*.xml` files and for those, fomats the output accordingly, using `<model>` element. However there are fixed lense entries outside of `compact*.xml`.

This can be fixed by finding and moving all fixed lense entries to `compact*xml` (not all fixed lense cameras are compact cameras though).

Alternatively, lenslist script can be changed to identify fixed lenses not by their placement in specifically named XML files, but by `model lang=en` being ‘fixed lens’ and then composing the name in lenslist with a particular model from `model` element. This should not require changes to the database. Lens Coverage then woul look like in the attached file:

[lensfun_coverage.html.txt](https://github.com/user-attachments/files/22083403/lensfun_coverage.html.txt)

Thoughts?

### DIFF
--- a/tools/lenslist/show_lensfun_coverage.py
+++ b/tools/lenslist/show_lensfun_coverage.py
@@ -47,7 +47,7 @@ class Lens:
     def __init__(self, element, root, camtype):
         self.maker = find_best(element, "maker")
         self.model = find_best(element, "model")
-        if camtype == "compact":
+        if self.model.startswith("fixed lens"):
             mount = element.find("mount").text
             for camera in root.findall("camera"):
                 if camera.find("mount").text == mount:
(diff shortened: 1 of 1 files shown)


Andries-B:
The lenslist looks much better now, because now there are no "anonymous" lenses anymore (like the GoPro's before). Also, I would not prefer to use a file-name or a specific file such as "compact.xml" as meta-data about what is in that file. It's better to have that information in the file, like in this solution: it is in the lang="en" tag. This is a practical solution that does not require a database change. Alternative might be to add another xml element, but that requires a database change.
I think this is a meaningful improvement.

tuxfanx:
The change would improve the lists. May be there are some special cases that need to be reviewed and possibly adjusted. For example, not all compact camera lens profiles contain the label "fixed lens" in the display name. These might need to be adjusted slightly. For example, in the compact-fujifilm.xml file, the X100 telephoto lens TCL-X100 or the profile for the X100V/VI, in which the fixed lens is combined with the wide-angle and tele lens converters.

PR #2577, opened 2025-09-12
Added all available mounts to Sigma 12-24 f/4 lens

thesolution90:
The Sigma 12-24 f/4 Art Lens is also available for Canon EF. The attached image can be used for testing the existing calibration data with the lens using a Canon EF mount.

[IMG_0012.zip](https://github.com/user-attachments/files/22297909/IMG_0012.zip)

### DIFF
--- a/data/db/slr-sigma.xml
+++ b/data/db/slr-sigma.xml
@@ -257,6 +257,11 @@
         <maker>Sigma</maker>
         <model>Sigma 12-24mm F4 DG HSM | A</model>
         <mount>Nikon F AF</mount>
+        <mount>Sigma SA</mount>
+        <mount>Canon EF</mount>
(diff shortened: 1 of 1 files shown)


tuxfanx (changes requested):
[According to Sigma](https://www.sigma-global.com/en/lenses/a016_12_24_4/), the lens was only produced with the mounts for Nikon F, Canon EF and Sigma SA. Therefore, only these mounts should be included in the profile. Options via adapters should not be included.
Users who use lens adapters can either [use the lensfun‑add‑adapter](https://lensfun.github.io/manual/v0.3.2/lensfun-add-adapter.html) script or can make the corresponding changes in their own database.

tuxfanx:
Added mounts for Canon EF and Sigma SA --> Closed

PR #2639, opened 2025-11-19
Add flag to reproduce distortion correction scaling of legacy Lensfun versions

seebk:
The internal coordinate systems in Lensfun have bee unified in `master` which leads to differently scaled outputs when using the `LF_DIST_MODEL_POLY3` and `LF_DIST_MODEL_PTLENS` models for distortion correction.

This creates problems with backwards compatibility and repoducibility: images from Darktable and other RAW processors will show a different scaling when being processed with Darktable using Lensfun < 0.3.x and Lensfun from `master`.  A bug report was already submitted (#945) but its description is a bit misleading. It's not the general scale factor that has changed but the scaling of some distortion models. A previous PR #1174 to address that issue was submitted but not merged after discussion in the comments.

Here’s another attempt to fix the issue, one that is closer to addressing the actual root cause. The `EnableDistortionCorrect()` method now includes a new flag that preserves scaling after distortion correction, making the output more consistent with that of previously released legacy Lensfun versions. Internally, enabling this flag adds an additional scaling callback to compensate for the differing output scales of the `LF_DIST_MODEL_POLY3` and `LF_DIST_MODEL_PTLENS` models.

### DIFF
--- a/include/lensfun/lensfun.h.in
+++ b/include/lensfun/lensfun.h.in
@@ -2148,14 +2148,14 @@ struct LF_EXPORT lfModifier
      * @return
      *     A set of LF_MODIFY_XXX flags in effect.
      */
-    int EnableDistortionCorrection (const lfLensCalibDistortion& lcd);
+    int EnableDistortionCorrection (const lfLensCalibDistortion& lcd, bool legacyScaling = false);
     /**
      * @brief Enable distortion correction.
      *
      * @return
      *     A set of LF_MODIFY_XXX flags in effect.
      */
-    int EnableDistortionCorrection ();
+    int EnableDistortionCorrection (bool legacyScaling = false);
 
     /**
      * @brief Enable TCA correction.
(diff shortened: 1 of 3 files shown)


seebk:
@bronger If you're still around I'd be curious to hear your thoughts on above solution. Does that make sense?

bronger:
I am still around but I would have to re-learn a lot of stuff. Can you give me a summary what the new coordinate system is, and its motivation?

Am I right that it is about incompatibilities introduced by myself?

seebk:
Yes, it's related to your edits many years ago and I hoped you better recall the details than I do :) 

As far as I understand it, the d component in the POLY3 and PT models was factored out to unify the internal coordinate systems for all models. See comments in modifier.cpp and mod-coord.cpp
https://github.com/lensfun/lensfun/blob/b2c60962aa6dc5ecfe99eefb124e9c90cb916750/libs/lensfun/modifier.cpp#L17
https://github.com/lensfun/lensfun/blob/b2c60962aa6dc5ecfe99eefb124e9c90cb916750/libs/lensfun/mod-coord.cpp#L10

Issue #2648, opened 2025-12-02
Olympus / OM System Branding Update


ghost:
Hello,

I am a RawTherapee and OM System 12-100 mm F4.0 user. I notice that RawTherapee could not find the right lens for my photo. That was because I used the re-branded lens. I fixed the issue by copying the [Olympus lens](https://github.com/lensfun/lensfun/blob/master/data/db/mil-olympus.xml#L1060-L1151) to a new OM System lens [here](https://github.com/lensfun/lensfun/blob/master/data/db/om-system.xml). I replaced the maker from "Olympus" to "OM Digital Solutions" and the model from "Olympus M.Zuiko Digital ED 12-100mm f/4.0 IS Pro" to "OM 12-100mm F4.0". I contacted OM Digital Solutions to make sure the change is only on the name. The support team confirmed that:

> there is no difference in optical performance - including distortion, chromatic aberration, or vignetting - between the Olympus-branded lens and the OM System-branded lens. Both versions are identical in design and functionality, featuring the same optical elements and construction. The only change is the branding engraved on the exterior, which now reflects OM System. The variation you notice in the EXIF metadata (e.g., “OM 12-100 mm F4.0” versus “Olympus 12-100 mm F4.0”) is simply due to this branding update. The optical characteristics remain unchanged.

If that sounds good, I could open a pull request to update [this file](https://github.com/lensfun/lensfun/blob/master/data/db/om-system.xml).

Best Regards,
Friedrich

tuxfanx:
Hi Friedrich,
please create a PR.

PR #2649, opened 2025-12-03
Add OM 12-100mm F4.0

ghost:
https://github.com/lensfun/lensfun/issues/2648


sarunasb:
I would suggest adding `lang="en"` lines, following the pattern from other entries `om-system.xml`.

PR #2669, opened 2026-01-20
Add vignetting data for Canon EF-S 18-135mm f/3.5-5.6 IS

jtniehof:
This PR adds vignetting data for the Canon EF-S 18-135mm f/3.5-5.6 IS. This is one of the kit lenses for the EOS 550D / Rebel T2i.

I've been sitting on this for a _long_ time, mostly hoping somebody else would do a better job and spare me the pain. I only have one exposure at each focal length / aperture / focus distance. This was of a white ceiling shot through a white gel, illuminated about as evenly as I could make it--not perfect, but I do think it's a lot better than not correcting for vignetting.

I did a lot of playing around with various fit approaches and I didn't really find anything better than the calibrate.py script.

If you're interested in seeing the before/after, [the vignetting images are here](https://drive.google.com/drive/folders/1NsFicxPdcTkPdTOMgu7_PZBZCyuPQvpK?usp=sharing). The average intensity was taken from the centre 2/3 in x and y (so from 1/6 to 5/6, i.e. 4/9 of the image area) and each pixel compared to that. White is a match--so a perfectly flat image would be all white. Saturated red is 1/4 stop above flat (i.e., 1.189x the average); saturated blue is 1/4 stop below. Color scale is log.


jtniehof:
Looks like I still have commit access...

I've started using this data with my images and I'm pretty happy with it--there's a noticeable difference and the illumination is much more even with the correction.

Unless someone objects otherwise, I'll merge this on January 29 (a week from today). Happy to provide any data that are requested.

jtniehof:
Requested some reviews just to make clear I'm not trying to slip this in quietly. Consider it an "FYI" rather than an assignment. Thanks.

jtniehof:
Hmm, I hadn't noticed the center moving, but you're right. Good catch. I also did take this with the camera pointing up at the ceiling, so the optics may shift differently than in normal usage. I suppose the "right" thing to do would be to take images with the lens horizontal and the camera in either landscape or portrait orientations, but the thought of not only doing this again but twice over fills me with dread :) As you say, it might not be possible to get better out of this lens.

jtniehof:
Yes, I've checked "real world" images with these corrections.

Issue #2647, opened 2025-12-02
Calibration upload dff1c6


bronger:
Calibration images were uploaded to the directory that starts with “`dff1c6_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> This is extra data for issue 4086ef #2621
> 
> This contains:
>  * More distortion data, specifically 55
>  * Vignetting using solid plastic under a cloudy sky
>  * Vignetting test photos

tuxfanx:
Your new photos for the distortion profile cannot be used by us. We use scripts for the profile generation that require photos to create the distortion profile, as can be read on several lensfun websites. Please read the following instructions and only submit distortion pics that meet the requirements.

For distortion calibration two continuous straight lines from the left to the right edge of the image must be contained in the picture. The 1st line very near to the top or bottom of the image and the 2nd line about 30% of the total image height away from the 1st line. The lines of the subject should nearly run parallel to the top or bottom edge of the image and they should have good contrast and sharpness.

How to take the pictures is described [here ](https://pixls.us/articles/create-lens-calibration-data-for-lensfun/)(section "Taking pictures") and partly also [here](https://wilson.bronger.org/lens_calibration_tutorial/), [here](https://wilson.bronger.org/calibration) and [here](https://wilson.bronger.org/calibration/target_tips).
If you upload more pictures please refer to this issue in the comment

Issue #2621, opened 2025-10-27
Calibration upload 4086ef


bronger:
Calibration images were uploaded to the directory that starts with “`4086ef_`”.

Please read the [workflow description](https://github.com/lensfun/lensfun/blob/master/tools/calibration_webserver/workflow.rst) for further instructions about the calibration.

Additional comments by the uploader:

> Lens: Nikon AF-S DX Nikkor 55-200mm f/4-5.6G ED VR II
> Camera: Nikon D90
> 
> I was struggling to find buildings wide enough, I can retake them if necessary
> No vignetting here (too big), just TCA & distortion

byteit101:
I am the uploader.  I was struggling to find buildings wide enough at the wide end while being far enough away to not be looking upwards too much. Let me know and I can retake them.

I also have Vignetting photos for this lens, but they total 1GB and I got a timeout. I did manage to run the vignetting script to get this output, but I'm not confident that the vignettting photos are good quality


```xml
            <vignetting model="pa" focal="55.0" aperture="4.0" distance="1.1" k1="-0.0388692" k2="0.0641616" k3="-0.2730264" />
            <vignetting model="pa" focal="55.0" aperture="4.0" distance="18.0" k1="-0.0172125" k2="-0.7132483" k3="0.3668198" />
            <vignetting model="pa" focal="55.0" aperture="4.5" distance="18.0" k1="0.0281204" k2="-0.7682960" k3="0.3865567" />
            <vignetting model="pa" focal="55.0" aperture="5.0" distance="18.0" k1="0.1163091" k2="-0.7606754" k3="0.3332967" />
            <vignetting model="pa" focal="55.0" aperture="5.6" distance="1.1" k1="-0.1351101" k2="0.3023142" k3="-0.4020132" />
            <vignetting model="pa" focal="55.0" aperture="5.6" distance="18.0" k1="-0.0871928" k2="0.0746525" k3="-0.1444462" />
            <vignetting model="pa" focal="55.0" aperture="6.3" distance="1.1" k1="-0.1756833" k2="0.3863485" k3="-0.4339865" />
            <vignetting model="pa" focal="55.0" aperture="6.3" distance="18.0" k1="-0.1130931" k2="0.1044461" k3="-0.1356383" />
            <vignetting model="pa" focal="55.0" aperture="7.1" distance="18.0" k1="0.0752288" k2="-0.3914408" k3="0.2218319" />
[...]

PR #2681, opened 2026-01-29
Add distortion profile for Nikkor AF-S 55-200mm f/4-5.6G DX ED VR II

byteit101:
Related to https://github.com/lensfun/lensfun/issues/2621 and https://github.com/lensfun/lensfun/issues/2647

I don't understand why the coefficients are swinging back and forth from positive to negative, but they seem to be ok on the images I've tested


byteit101:
Ah oops, I was using the hugin default projection, and not rectilinear. I have re-generate this and the other profile that I attempted distortion for

PR #2683, opened 2026-01-29
Add full-frame lens profile for Canon EF 75-300 III

byteit101:
This uses Bronger's calibrate.py for vignetting (sha256sum = cab4ac0dd7b6986ce281906306d704f5960e8d1e5378d0c5e2163ab1da4fa99c)

TCA was a manually-selected average of both automatic and manually-calibrated points

### DIFF
--- a/data/db/slr-canon.xml
+++ b/data/db/slr-canon.xml
@@ -8045,11 +8045,11 @@
         <cropfactor>1.005</cropfactor>
         <calibration>
             <!-- Taken with Canon 6D -->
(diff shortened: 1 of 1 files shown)


tuxfanx:
Thanks for the test pictures.
The vignetting correction works very well. But there seems to be something wrong with the distortion correction. This is particularly visible on the 75mm test image with the board wall. Please check it.

byteit101:
Ah oops, I was using the hugin default projection, and not rectilinear. I have re-generate this and the other profile that I attempted distortion for

PR #2687, opened 2026-02-02
Modernize calibration tool

yuri1969:
Recently, I've been using the tool for the first time. This made me spend some time figuring out how to use it correctly...

---

Here goes a round of modernization potentially improving the readability of the script. Feel free to suggest a breakdown of the PR to a set of smaller ones.

Testing was performed by simply comparing the generated `lenfun.xml` file.

---

Note Python 3.9 is now required.

- Replaced hand-crafted XML string concatenation
- Using f-strings now
- Replaced codecs.open() with the builtin
- Added explicit UTF-8 encoding
- Replaced communicate() with subprocess.run()
- Used context managers for Pool() and open()

### DIFF
--- a/tools/calibrate/calibrate.py
+++ b/tools/calibrate/calibrate.py
@@ -2,9 +2,10 @@
 #
 # Copyright 2012-2016 Torsten Bronger <bronger@physik.rwth-aachen.de>
 
-missing_packages = set()
+import subprocess, os, os.path, sys, multiprocessing, math, re, contextlib, glob, struct, argparse
+import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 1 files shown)


christopher-kapic:
Hi @yuri1969! I just made a CLI in Rust that compares the JPEG preview of a raw file to the raw file and identifies points to match and determine the best distortion values, with more standard means of calculating vignette/tca. Feel free to fork/copy this project and take it over, as I don't really care that much to maintain it. It works right now, but I think maybe the vignette could use some fine-tuning and the user experience could be better. Maybe you could compile it to WASM and build a simple web app to generate the XML in a user-friendly way?

Good luck!

You can find it [here](https://github.com/christopher-kapic/lensfun-generator).

PR #2715, opened 2026-02-19
Add vignetting and EF mount for Samyang/Rokinon 14mm

byteit101:
I kept it in mil-samyang as thats where it was, but as this is the EF mount, I suppose I could move it to the slr-samyang page istead


tuxfanx:
Isn't the lens also available with a Nikon F mount? If so, please add it. 
Please remove the blank line below distortion.

PR #2716, opened 2026-02-19
Add Tokina 116 II with Vignetting and TCA

byteit101:
Note that optically it should be the same as 116, but I have found that the existing vignetting profile is too strong for the photos that I've taken. I wasn't quite sure what to do.

### DIFF
--- a/data/db/slr-tokina.xml
+++ b/data/db/slr-tokina.xml
@@ -121,6 +121,65 @@
 
     <lens>
         <maker>Tokina</maker>
+        <model>Tokina 11-16mm f/2.8 AT-X 116 AF Pro DX II</model>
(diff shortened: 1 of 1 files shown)


tuxfanx:
Could you provide us with some sample raw pictures?

tuxfanx:
@byteit101 
Thanks for the pics in #2738
Since the test photos were taken from a very short distance, the profile shows a slightly too strong correction on some images, but this is still OK. Why don't you add distortion correction? You could also take the distortion data from the optically identical lens of version I. 
Please remove the blank lines in the calibration section.

tuxfanx:
The profile was adopted manually with a few changes.

byteit101:
Ah, thank you! I didn't add distortion correction before because I wasn't sure if I should.

PR #2736, opened 2026-03-02
Add Sony 24-50 f/2.8 G

christopher-kapic:
I vibe-coded a Rust CLI which compares Sony (and maybe other camera models) JPEG previews against the raw files to calculate distortion. You can find it [here](https://github.com/christopher-kapic/lensfun-generator).

Using this, I generated the values for the 24-50 f/2.8 G. I tested it in RapidRaw, and it seems to work correctly.

### DIFF
--- a/data/db/mil-fujifilm.xml
+++ b/data/db/mil-fujifilm.xml
@@ -2199,6 +2199,24 @@
         </calibration>
     </lens>
 
+	<lens>
+        <maker>Fujifilm</maker>
+        <model>GF35-70mmF4.5-5.6 WR</model>
+        <mount>Fujifilm G</mount>
+        <cropfactor>0.79</cropfactor>
+        <calibration>
+            <!-- Taken with Fujifilm GFX50S -->
+            <distortion model="ptlens" focal="35" a="0.0139605" b="-0.0519035" c="0.0110288"/>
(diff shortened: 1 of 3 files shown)


christopher-kapic:
@tuxfanx is there anything else you need me to do to make this ready to merge? I have tested with my camera and RapidRaw, and these values seem to be good. I can upload some sample images if that is necessary.

tuxfanx:
The manufacturers do not always fully correct the aberrations of their lenses with the corresponding correction profiles. This is particularly true for vignetting. Your profile may therefore not correspond to what the lensfun scripts would generate. 
I have only limited programming knowledge and cannot judge whether your generated profile is suitable for a merge without test images. For my part, I generally only accept profiles if I have been able to at least roughly test them using test images or have already had positive experience with the uploader.  Since you asked me personally, yes, I need test images. [You can provide them](https://wilson.bronger.org/calibration) here. Make sure that they also include pictures taken with the aperture open. Six to eight meaningful pictures should be sufficient.

I tried to test your profile [with this raw image](https://www.dpreview.com/sample-galleries/4349107183/sony-24-50mm-f2-8-g-sample-gallery/2841753257) from the Internet using darktable. The distortion correction did not work at all with this image.

christopher-kapic:
Great, thank you. I'll work on that. I appreciate your work and response.

PR #2768, opened 2026-04-22
Add a simple table-filter function for the generated lens list HTML

byteit101:
This adds an integrated search box to the generated lens list. I made some choises that can be revisited, but it is handy so far in the current state. I may also have missed some key things, but it seems to work fine when I generate it locally.

Notable things:
- Done entirely in a progressive-enhancement way with client-side JavaScript.
- Negatives and case-insensitivity are supported
- Searching on crop factor is not supported
- Regex and fancy searches are not supported
- Right now, the entire script is embedded in the output HTML
- The search box isn't styled and is fairly hard-coded

Screenshots of various searches:
Screenshot_20260421_225207
Screenshot_20260421_224929


tuxfanx:
@bronger @sarunasb  
A filter function would be helpful.

PR #2780, opened 2026-04-28
Add Sigma 150mm f/2.8 EX DG APO HSM Macro for APS-C

byteit101:
This PR adds a 1.5 crop to the 1.0 and 2.0 crop verions of the Sigma 150mm EX Macro lens.

Taken with a Nikon D90. Vignetting test photos will be uploaded an linked a few minutes after I post this PR. Edit: #2782 

I also have a 1.4x TC version of this, but I was encountering issues, see this issue: #2784


tuxfanx:
There is already a profile for crop 1.0. It can also be used for crop 1.53. But since your profile seems to be a bit better --> merge

Issue #2811, opened 2026-06-08
Sony ICLE-6400A and ICLE-6100A


mcourteaux:
New models which change the charging of the battery, but are otherwise identical to the ICLE-6400 and ICLE-6100 models, were introduced a year ago:

https://thenewcamera.com/sony-silently-announces-the-sony-a6400a-camera-as-well-as-the-sony-a6100a/
https://www.reddit.com/r/SonyAlpha/comments/1hyudc8/new_model_ilce6400a/

I don't know if we want to list them as just alternative names of the same camera, or we add it as a new camera?

tuxfanx:
It depends on model names in the exif data. If the names are different we have to add them as new cameras.

mcourteaux:
Ah yes, I forgot to specify. Yes they effectively read out the following:

<img width="650" height="137" alt="Image" src="https://github.com/user-attachments/assets/bc6c96b3-3363-44a8-b423-ff9d82b93fab" />

I don't have the 6100 camera, so I can't test it, but I'd assume it's the same.

(If interested, the EXIF reader I'm making is https://github.com/mcourteaux/neonEXIF)

kmilos:
Btw, same goes for ILME-FX3A and ZV-1A.

tuxfanx:
Unfortunately, I do not have any reliable exif data for the other cameras mentioned above.

kmilos:
The 6400A sample can be found on [RPU](https://raw.pixls.us/).

tuxfanx:
@kmilos 
RX100III: 8.8-25.7mm
ZV-1: 9.4-25.7mm

PR #2814, opened 2026-06-12
Normalize 7RM4A pretty name

kmilos:
Make it consistent w/ all other "A" variants.

See also https://github.com/lensfun/lensfun/issues/2811


kmilos:
Re failure: Windows GHA runners are currently in the process of transitioning to VS 2026, removing the -G option (let CMake use the default) should fix it: https://github.com/lensfun/lensfun/blob/719e3b2a2a80b36684e52d5e791862bc6f1d0826/.github/workflows/cmake-windows.yml#L35

PR #2841, opened 2026-08-08
Added an alias for newer revision of Minolta AF 50mm f1.4

l00z:
Hey,

I've got a revision of this lens which registers as "Minolta AF 50mm F1.4 [New]" in EXIF data - this simply adds an alias since it should be optically identical to the original.

Sample photo attached.
[DSC05631.zip](https://github.com/user-attachments/files/30857210/DSC05631.zip)

### DIFF
--- a/data/db/slr-konica-minolta.xml
+++ b/data/db/slr-konica-minolta.xml
@@ -494,6 +494,7 @@
     <lens>
         <maker>Minolta</maker>
         <model>Minolta AF 50mm f/1.4</model>
+        <model>Minolta AF 50mm F1.4 [New]</model>
         <mount>Minolta AF</mount>
         <cropfactor>1</cropfactor>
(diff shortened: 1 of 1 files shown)


tuxfanx:
As far as I know, two model lines are uncommon and should not be used.
Either we add 2 separate profiles or just leave only the new model line. As far as I know, the detection of the old lens should also work with the new one. Unfortunately I don't have any images taken with the old lens available for testing.

l00z:
Yep, it's rare, only nikon has it:
```
$ head -6055 slr-nikon.xml | tail -2
        <model>Nikon AF-P DX Nikkor 10-20mm f/4.5-5.6G VR 168</model>
        <model>Nikon AF-P DX Nikkor 10-20mm f/4.5-5.6G VR</model>
```

